PR #40, opened 2026-02-02
Fix: audio export path resolution

Spyabo:
## Summary
- Fixed audio export failing when Tauri save dialog returns object instead of string path
- Added validation to prevent writing with invalid paths
- Changed from writeBinaryFile to writeFile API

## Test plan
- [x] Built Windows installer
- [x] Tested audio export from both Generate and Story tabs
- [x] Verified files save correctly to chosen location

## Note
**Needs macOS testing** - Please verify this doesn't break file saving on macOS before merging.


jamiepine:
sorry if I merged that too soon haha

PR #89, opened 2026-02-18
Linux Support

rocopolas:


backend/main.py
```diff
@@ -20,9 +20,20 @@
 from pathlib import Path
 import uuid
 import asyncio
-import signal
 import os
 
+# Set HSA_OVERRIDE_GFX_VERSION for AMD GPUs that aren't officially listed in ROCm
+# (e.g., RX 6600 is gfx1032 which maps to gfx1030 target)
+# This must be set BEFORE any torch.cuda calls
+if not os.environ.get("HSA_OVERRIDE_GFX_VERSION"):
```
mikeswann, line 28: I believe this variable is never set, even with ROCm listed GPUs and therefore this would set an incorrect override, causing possible issues for different cards. It would probably need to be mentioned in docs for eventual linux instructions, or added as setting in the app.

jamiepine:
Closing in favor of #262 which implements full Linux support including PulseAudio/PipeWire audio capture, AMD ROCm GPU detection, Whisper Turbo model support, and Linux CI. Thank you @rocopolas for kicking this off!

PR #93, opened 2026-02-18
fix(mlx): bundle native libs and broaden error handling for Apple Silicon

iJaack:
## Problem

The distributed macOS aarch64 binary (`voicebox_aarch64.app.tar.gz`) silently falls back to PyTorch CPU instead of using MLX, despite the README advertising 4-5x faster inference on Apple Silicon.

Two root causes:

### 1. `platform_detect.py` only catches `ImportError`

Inside a PyInstaller bundle, MLX loads its Metal shader libraries (`.metallib`) at import time. When those files are missing from the bundle, Python raises `OSError` — not `ImportError`. The original code:

```python
try:
    import mlx
    return "mlx"
except ImportError:   # ← OSError slips through here
    return "pytorch"
```

This causes a silent fallback to PyTorch on every Apple Silicon machine running the binary.

### 2. `collect_data_files` misses native libraries

`build_binary.py` and `voicebox-server.spec` used `--collect-data` / `collect_data_files` for `mlx` and `mlx_audio`. This only copies pure-Python files — native `.dylib` and `.metallib` files are excluded. PyInstaller needs `--collect-all` / `collect_all` + `Analysis(binaries=...)` to bundle shared libraries correctly.

## Fix

1. **`platform_detect.py`** — catch `(ImportError, OSError, RuntimeError)` and import `mlx.core` (forces native lib loading eagerly so any failure is caught here)
2. **`build_binary.py`** — replace `--collect-data mlx/mlx_audio` with `--collect-all`
3. **`voicebox-server.spec`** — use `collect_all()` and pass collected binaries to `Analysis(binaries=...)`

## Testing

Reproduced on macOS 15 (Apple Silicon M4 Mac mini):
- **Before fix**: binary logs `Backend: pytorch`, generation takes 30+ min for a 5s clip
- **After fix (source build with MLX)**: logs `Backend: MLX | GPU: MPS (Apple Silicon)`, same clip in ~39s (first run) / ~8s (warm)

### DIFF
--- a/backend/build_binary.py
+++ b/backend/build_binary.py
@@ -83,9 +83,13 @@ def build_server():
             '--hidden-import', 'mlx_audio.stt',
             '--collect-submodules', 'mlx',
             '--collect-submodules', 'mlx_audio',
-            # Collect MLX data files including Metal shader libraries (.metallib)
-            '--collect-data', 'mlx',
-            '--collect-data', 'mlx_audio',
+            # Use --collect-all so PyInstaller bundles both data files AND
+            # native shared libraries (.dylib, .metallib) for MLX.
+            # Previously only --collect-data was used, which caused MLX to
+            # raise OSError at runtime inside the bundled binary because
+            # the Metal shader libraries were missing.
+            '--collect-all', 'mlx',
+            '--collect-all', 'mlx_audio',
         ])
     else:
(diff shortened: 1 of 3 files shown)


Gomesproga:
before and binary count and servers e input in the lag das

Gomesproga:
woo , the apply is very good , an terminate and addons corresct the process

PR #97, opened 2026-02-18
fix: pass language parameter to TTS models

jannone:
## Summary

- **Both backends silently dropped the `language` parameter** — it was accepted by `generate()` but never forwarded to the underlying Qwen3-TTS model, causing it to default to "Auto" detection which frequently confuses Portuguese for Spanish
- **Added `LANGUAGE_CODE_TO_NAME` mapping** (ISO 639-1 → full language name) to both PyTorch and MLX backends, matching the format Qwen3-TTS expects (e.g. `"pt"` → `"portuguese"`)
- **PyTorch backend**: passes `language=` to `generate_voice_clone()`
- **MLX backend**: passes `lang_code=` to all 4 `model.generate()` call sites
- **Frontend**: auto-syncs the generation form's language dropdown with the selected voice profile's language via a new `useEffect`

## Test plan

- [ ] Select a Portuguese voice profile → language dropdown should auto-sync to Portuguese
- [ ] Generate speech in Portuguese → output should sound Portuguese, not Spanish
- [ ] Switch to English profile → language dropdown syncs to English, generation works normally
- [ ] Test with other supported languages (Chinese, Japanese, etc.)
- [ ] Check backend logs for any language-related errors

### DIFF
--- a/app/src/components/Generation/FloatingGenerateBox.tsx
+++ b/app/src/components/Generation/FloatingGenerateBox.tsx
@@ -13,7 +13,7 @@ import {
 } from '@/components/ui/select';
 import { Textarea } from '@/components/ui/textarea';
 import { useToast } from '@/components/ui/use-toast';
-import { LANGUAGE_OPTIONS } from '@/lib/constants/languages';
+import { LANGUAGE_OPTIONS, type LanguageCode } from '@/lib/constants/languages';
 import { useGenerationForm } from '@/lib/hooks/useGenerationForm';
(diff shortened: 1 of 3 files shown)


jamiepine:
Fixed on main in ca74c15 — both backends were silently dropping the language parameter. Applied the same approach from this PR: `LANGUAGE_CODE_TO_NAME` mapping + forwarding to both PyTorch (`language=`) and MLX (`lang_code=`) backends, plus the frontend profile-language sync. This was indeed the root cause of the language complaints. Thanks for the original diagnosis!

Issue #127, opened 2026-02-20
not able to find my gpu


kritik-sah:
<img width="1913" height="1011" alt="Image" src="https://github.com/user-attachments/assets/300ba208-8e9c-44cf-860d-4f7d7efc607e" />

sylvanmiori:
Same issue here on Apple Silicon Mac (arm64).

**Environment:**
- Mac mini (Apple Silicon, arm64)
- macOS (Darwin)
- Voicebox v0.1.0 (aarch64 build)

**Symptoms:**
-  in /health endpoint
- Model loading fails with: "Cannot copy out of meta tensor; no data! Please use torch.nn.Module.to_empty() instead of torch.nn.Module.to() when moving module from meta to a different device."
- Falls back to PyTorch CPU path instead of MLX/Metal

**What I've tried:**
1. Launched Voicebox.app directly (not CLI server)
2. Both 0.6B and 1.7B models downloaded successfully (~1.7GB and ~3.6GB)
3. API works for profile creation/sample upload
4. Generation fails with meta tensor error

**Root cause hypothesis:**
The GPU detection isn't recognizing Apple Silicon Metal/MLX properly, so it falls through to PyTorch CPU inference which has the accelerate meta tensor issue.

**Expected behavior:**
Should use MLX backend with native Metal acceleration on Apple Silicon.

Happy to help test any fixes!

kritik-sah:
fix it bro don't ask for permission

Cariaga:
same in my 1060 gpu built from source too.

Archviz360:
same here 3060

helpmebro911:
+1
nvidia 5090

mtpc4s9:
<img width="1742" height="788" alt="Image" src="https://github.com/user-attachments/assets/dee7b86e-1f83-4c3e-ab8d-f225b8e8d124" />

RTX 5080 had the issue: GPU not available

Anyone please support

dr3adl0rd:
RTX 2060 S, same problem, not finding GPU.

DE4DLY:
Same here

- RTX 5090

Ryderjj89:
Lol why is this referenced in the latest version as something being fixed when it's not? Latest version on WINDOWS still can't detect my RTX 3060.

gaurav1089:
I have RTX 4060 and GPU is not getting detected. But, in the latest commit its mentioned that its been resolved.

donn124:
[...]

Issue #125, opened 2026-02-20
How can I change default download path


amjadjibon:
I want to use different storage to download the models. How can I change the default download path?

amjadjibon:
Thanks for resolving this

Issue #121, opened 2026-02-20
[Bug] Web version missing @tailwindcss/vite plugin - CSS not loading


white1107:
## Description
The web version (`web/vite.config.ts`) is missing the `@tailwindcss/vite` plugin, causing CSS to not load at all. The page renders without any styles, appearing as a blank white page.

## Steps to Reproduce
1. Clone the repository
2. Run `make setup`
3. Run `bun run dev:web`
4. Open http://localhost:5173 in browser
5. Observe: No CSS is applied, UI appears broken/white

## Affected Browsers
- Chrome (tested)
- Safari (tested)

## Expected Behavior
CSS should load correctly, matching the Tauri desktop version.

## Root Cause
`tauri/vite.config.ts` includes the Tailwind plugin:
```ts
import tailwindcss from '@tailwindcss/vite';
// ...
plugins: [react(), tailwindcss()],
```

But `web/vite.config.ts` is missing it:
```ts
plugins: [react()],  // Missing tailwindcss()
```

## Proposed Fix
Add `@tailwindcss/vite` to `web/vite.config.ts` and install the dependency.

## Environment
- macOS (Apple Silicon)
- Tailwind CSS v4
- Vite v5.4.21

Issue #119, opened 2026-02-20
Not Support Interl Arc GPU


sunjiawe:
My GPU model: Intel(R) Arc(TM) 130T GPU  
OS: Windows 11  

In Voice Box, the Server Status shows: "GPU: Not Available"

PR #128, opened 2026-02-20
fix: resolve multiple issues (#96, #119, #111, #108, #121, #125, #127)

mrigankad:
This PR addresses several critical bugs and feature requests identified in recent sessions, ranging from model loading logic to hardware support and UI/UX improvements.

#### Round 1 — Infrastructure & Compatibility
* **Issue #121 — CSS Missing in Web Version**: Added Tailwind CSS Vite plugin to web/vite.config.ts and updated dependencies.
* **Issue #127 — "Cannot copy out of meta tensor"**: Fixed CPU fallback on Apple Silicon by bypassing meta-tensor initialization when using CPU.
* **Issue #125 — Custom Model Directory**: Implemented VOICEBOX_MODELS_DIR environment variable support in backend/config.py.

#### Round 2 — Recording & Upload Stability (Issue #108)
* **Bug A (MP3 Uploads)**: Fixed backend crash when uploading non-WAV files by correctly extracting file extensions.
* **Bug B (Duration Mismatch)**: Replaced fragile <audio> duration reading with AudioContext.decodeAudioData for accurate MP3 length detection.
* **Bug C (Conversion)**: Ensured all manually uploaded samples are converted to WAV on the client side before submission.
* **Bug D (Recording Race Condition)**: Fixed a bug where cancelling a recording would still trigger a "complete" event with an empty blob.

#### Round 3 — Performance & Hardware (Issues #96, #119, #111)
* **Issue #96 — 0.6B Model Selection**: Fixed race condition where the 1.7B default model was loaded during voice prompt creation before the 0.6B model was initialized.
* **Issue #119 — Intel Arc / DirectML Support**: Added hardware detection for Intel XPU (IPEX) and DirectML (torch-directml) backends.
* **Issue #111 — Audio Streaming**: Added a new /generate/stream endpoint for instant playback without intermediate disk writes.

### DIFF
--- a/app/src/components/VoiceProfiles/ProfileForm.tsx
+++ b/app/src/components/VoiceProfiles/ProfileForm.tsx
@@ -43,7 +43,7 @@ import {
 } from '@/lib/hooks/useProfiles';
 import { useSystemAudioCapture } from '@/lib/hooks/useSystemAudioCapture';
 import { useTranscription } from '@/lib/hooks/useTranscription';
-import { formatAudioDuration, getAudioDuration } from '@/lib/utils/audio';
+import { convertToWav, formatAudioDuration, getAudioDuration } from '@/lib/utils/audio';
 import { usePlatform } from '@/platform/PlatformContext';
 import { useServerStore } from '@/stores/serverStore';
 import { type ProfileFormDraft, useUIStore } from '@/stores/uiStore';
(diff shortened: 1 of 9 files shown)


Cdddo:
I see this was merged but I wish streaming would work inside the GUI and not just via the endpoint...

PR #267, opened 2026-03-13
MCP server for voicebox

l33t0:
Adds a Rust MCP server (`mcp/`) that lets spacebot and other MCP clients use voicebox as a tool provider.

## What this does

Thin proxy binary that sits between MCP clients and the voicebox FastAPI backend. Translates MCP tool calls into HTTP requests against `localhost:17493`.

28 tools across the full API surface:
- **Voice profiles** - CRUD, samples, cloning
- **Speech generation** - text-to-speech with voice profiles, audio download
- **History** - browse/search/stats for past generations
- **Transcription** - audio-to-text via Whisper
- **Stories** - multi-voice timeline projects, export
- **Models** - status, load/unload, download from HuggingFace
- **System** - health check, cache management

## Transport

Both stdio and streamable HTTP. Stdio is the default for subprocess-based clients like spacebot. HTTP for shared/remote access.

```bash
# stdio (default)
voicebox-mcp --url http://localhost:17493

# http
voicebox-mcp --transport http --port 3100
```

## Spacebot config

```toml
[[mcp_servers]]
name = "voicebox"
transport = "stdio"
command = "/path/to/voicebox-mcp"
```

## Stack

- Rust, rmcp 1.2 (same MCP SDK spacebot uses as client)
- reqwest for HTTP proxying
- clap for CLI, tracing for logging
- 18 tests via wiremock

## Test plan

- [x] `cargo fmt -- --check` clean
- [x] `cargo clippy -- -D warnings` clean
- [x] `cargo test` - 18/18 passing
- [ ] Manual test with spacebot stdio connection
- [ ] Manual test with HTTP transport

## Summary by CodeRabbit

* **New Features**
  * Added Voicebox MCP service with CLI (stdio/HTTP), HTTP endpoint, voice profiles, text-to-speech generation, audio download, transcription, stories, model management, history, and system health/cache controls.
* **Documentation**
  * New developer guide with quick-start, transport options, configuration, and usage examples.
* **Tests**
  * Added extensive unit and integration tests validating MCP endpoints and client interactions.
* **Chores**
  * Added container/Docker support and local build/run/test targets for the MCP service.


jamiepine:
Hey, thanks for putting this together -- the code quality is solid and the rmcp usage is clean.

That said, I think a standalone Rust binary that proxies HTTP to the FastAPI backend isn't the right layer for this. A few concerns:

- It duplicates the entire API surface as hand-written reqwest calls (~2k lines) that need to stay in sync with backend changes. Any time we add or modify an endpoint, this proxy needs a matching update.
- It doesn't integrate with Tauri at all -- it can't leverage server lifecycle management (start/stop/restart), so it just assumes the backend is already running somewhere.
- It introduces a separate Cargo workspace with its own dependency tree, which is extra maintenance overhead.

I think MCP support would fit better at the Python backend level. The FastAPI server already has all the endpoints defined, so we could register them as MCP tools directly alongside the existing REST routes using the Python MCP SDK. No code duplication, no extra process to manage, and it stays in sync with the API automatically.

Going to close this for now but appreciate the effort here.

PR #320, opened 2026-03-18
feat: Intel Arc (XPU) GPU support

jamiepine:
## Summary

Adds Intel Arc GPU (XPU) support so that users with Intel Arc A-series GPUs (e.g. A770, A750) can use hardware acceleration instead of falling back to CPU.

Prompted by user feedback: an Intel Arc A770 16GB owner installed `intel-extension-for-pytorch` system-wide, but voicebox's venv didn't include it, and most backends didn't check for XPU anyway.

## Changes

- **`justfile`** — Windows `setup-python` now auto-detects Intel Arc GPUs (via `Win32_VideoController`) and installs `torch` from the XPU wheel index + `intel-extension-for-pytorch`, mirroring the existing NVIDIA/CUDA auto-detection
- **`backend/backends/base.py`** — Added two shared helpers:
  - `empty_device_cache(device)` — clears VRAM on CUDA or XPU after model unload
  - `manual_seed(seed, device)` — sets reproducible seeds on CPU + CUDA or XPU
- **All TTS backends** now pass `allow_xpu=True` to `get_torch_device()`:
  - `chatterbox_backend.py`
  - `chatterbox_turbo_backend.py`
  - `hume_backend.py` (also enables bf16 on XPU — Intel Arc supports it natively)
  - `luxtts_backend.py`
  - `pytorch_backend.py` (already had `allow_xpu=True`, now uses shared helpers)
- Replaced all inline `torch.cuda.empty_cache()` / `torch.cuda.manual_seed()` calls with the shared helpers so XPU gets the same treatment

## Device priority

The existing priority in `get_torch_device()` is unchanged: **CUDA > XPU > DirectML > MPS > CPU**. Users with both NVIDIA and Intel GPUs will still use CUDA.

## Testing notes

- Requires an Intel Arc GPU + oneAPI toolchain to test XPU path end-to-end
- CPU and CUDA paths are unchanged (the helpers are no-ops for non-matching devices)
- The `setup-python` detection uses `Intel.*Arc` regex, so Intel integrated GPUs (UHD, Iris) won't trigger the XPU install

## Summary by CodeRabbit

* **New Features**
  * Added Intel Arc (XPU) detection and optional installation path; UI/backend now report XPU devices where available.
  * Health/status reporting now accounts for XPU memory usage when present.

* **Chores**
  * Unified device-aware cache clearing and seed initialization across backends for consistent behavior on CPU/CUDA/XPU.
  * Refined device selection logic and unload handling for more robust resource management.

### DIFF
--- a/backend/app.py
+++ b/backend/app.py
@@ -155,6 +155,20 @@ def _get_gpu_status() -> str:
         return "MPS (Apple Silicon)"
     elif backend_type == "mlx":
         return "Metal (Apple Silicon via MLX)"
+
+    # Intel XPU (Arc / Data Center) via IPEX
+    try:
+        import intel_extension_for_pytorch  # noqa: F401
+
+        if hasattr(torch, "xpu") and torch.xpu.is_available():
+            try:
+                xpu_name = torch.xpu.get_device_name(0)
+            except Exception:
+                xpu_name = "Intel GPU"
+            return f"XPU ({xpu_name})"
+    except ImportError:
+        pass
+
     return "None (CPU only)"
 
 
--- a/backend/backends/base.py
+++ b/backend/backends/base.py
@@ -126,6 +126,37 @@ def get_torch_device(
     return "cpu"
 
 
+def empty_device_cache(device: str) -> None:
+    """
+    Free cached memory on the given device (CUDA or XPU).
+
+    Backends should call this after unloading models so VRAM is returned
+    to the OS.
+    """
+    import torch
+
+    if device == "cuda" and torch.cuda.is_available():
+        torch.cuda.empty_cache()
+    elif device == "xpu" and hasattr(torch, "xpu"):
+        torch.xpu.empty_cache()
+
+
+def manual_seed(seed: int, device: str) -> None:
+    """
+    Set the random seed on both CPU and the active accelerator.
+
+    Covers CUDA and Intel XPU so that generation is reproducible
+    regardless of which GPU backend is in use.
+    """
+    import torch
+
+    torch.manual_seed(seed)
+    if device == "cuda" and torch.cuda.is_available():
+        torch.cuda.manual_seed(seed)
+    elif device == "xpu" and hasattr(torch, "xpu"):
+        torch.xpu.manual_seed(seed)
+
+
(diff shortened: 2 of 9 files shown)


oburtchen:
I checked it out, but unfortunately still no GPU acceleration. One thing I noted, in my system python installation, the intel pytorch extension was installed this way:

```
python -m pip install torch==2.8.0 torchvision==0.23.0 torchaudio==2.8.0 --index-url https://download.pytorch.org/whl/xpu
python -m pip install intel-extension-for-pytorch==2.8.10+xpu --index-url https://pytorch-extension.intel.com/release-whl/stable/xpu/us/
```
So the extension comes from intel servers. Don't know if this makes a difference. Okay, the compile log is attached, just tell me if I can help in any way.
[voicebox-compile-log.txt](https://github.com/user-attachments/files/26098313/voicebox-compile-log.txt)

jamiepine:
Thanks for testing this and providing the feedback! We're looking into the Intel Arc acceleration issue now. The compile log you attached will be helpful for debugging. Will update you once we identify what's causing the GPU acceleration to not work despite the code changes.

Zenvita:
# Voicebox PR #320 Troubleshooting Notes

## Environment
* **Hardware:** Intel Arc Pro B70 (Battlemage) / Intel Core Ultra 7 CPU
* **OS:** Linux Mint Cinnamon

## Issue Encountered
While model weight loading onto Intel XPU (`xpu`) succeeds, executing the token generation loop triggers a hard C++ assertion failure in PyTorch's nightly SYCL indexing and tensor comparison kernels:

```text
Index.h:622: Assertion `index >= -sizes_[i] && index < sizes_[i] && "index out of bounds"` failed.
TensorCompareKernels.cpp:180: Assertion `input_[0] != 0` failed.
```

This is an upstream kernel bug in PyTorch's advanced indexing or padding mask handling on the Battlemage architecture during autoregressive generation.

## Workaround / Resolution
Forcing the backend execution device to CPU (`self.device = "cpu"`) bypasses the_broken XPU indexing kernel entirely, providing stable audio generation without crashing the process.

Zenvita:
Thank you!!

PR #418, opened 2026-04-15
Add initial frontend quality gates and TypeScript hardening

erionjuniordeandrade-a11y:
Summary
- add a lightweight CI workflow that enforces typecheck + web build on PRs
- make root `bun run ci` meaningful and passing
- fix the TypeScript issues blocking app/web typecheck
- update stale version examples in SECURITY.md

Why
- the repo had release automation, but no PR-time quality gate
- frontend type errors were being worked around instead of surfaced
- stale security version examples were drifting behind current releases

Verification
- `bun run typecheck`
- `bun run build:web`
- `bun run ci`

Notes
- I intentionally did not gate on `biome check` yet because the repo still has substantial pre-existing formatting/a11y debt that would keep CI permanently red.
- This PR is the first hardening slice, not the full cleanup.


erionjuniordeandrade-a11y:
Follow-up tracking issues opened from this hardening pass:
- #419 Reduce engine sprawl: define first-class vs experimental TTS backends
- #420 Reduce platform sprawl: formalize tier-1 and tier-2 support targets
- #421 Track and burn down frontend Biome + accessibility debt before gating CI
- #422 Code-split the web build: current main bundle is over 1 MB

That gives the repo a cleaner path from “make CI meaningful” to “make architecture maintainable.”

erionjuniordeandrade-a11y:
I spun up the next slice as a stacked follow-up PR on my fork:
- https://github.com/erionjuniordeandrade-a11y/voicebox/pull/1

That branch introduces explicit engine/platform support tiers and lazy-loads the heavy routes. Local verification dropped the main web bundle from about 1069 kB to about 637 kB minified while keeping `bun run ci` green.

I kept it stacked on top of this hardening PR so the review surface stays sane instead of mixing architecture work into the first patch.

PR #429, opened 2026-04-16
fix: harden voice prompt cache loading and SPA path guard

shaun0927:
## Summary

- Add `weights_only=True` to the `torch.load()` call in the voice prompt cache
- Replace `str.startswith()` with `Path.is_relative_to()` in the SPA catch-all route

### 1. Voice prompt cache — `backend/utils/cache.py`

`get_cached_voice_prompt()` calls `torch.load(cache_file)` without `weights_only=True`. Since PyTorch 2.6 this triggers a `FutureWarning`, and the [PyTorch docs recommend](https://pytorch.org/docs/stable/generated/torch.load.html) explicitly opting in to the safe unpickler for all `torch.load()` calls.

The cached `.prompt` files contain dicts of tensors and basic Python types (strings, ints, floats), all of which are supported by the safe unpickler. No functional change is expected.

```python
# Before
prompt = torch.load(cache_file)

# After
prompt = torch.load(cache_file, weights_only=True)
```

### 2. SPA path guard — `backend/app.py`

The `serve_spa` catch-all route (active in Docker/web deployment when `frontend/` exists) guards against path traversal using:

```python
str(file_path).startswith(str(frontend_dir))
```

This string prefix comparison has a subtle flaw: if `frontend_dir` is `/app/frontend`, then a path like `/app/frontend_evil/secret` also passes the check because the string `"/app/frontend_evil"` starts with `"/app/frontend"`.

`Path.is_relative_to()` (Python 3.9+) correctly tests directory containment:

```python
# Before
str(file_path).startswith(str(frontend_dir))

# After
file_path.is_relative_to(frontend_dir)
```

Note: the existing `Path.resolve()` call already blocks `../` traversal. This fix addresses the narrower prefix-collision case.

## Verification

- `py_compile` passes on both modified files
- `torch.load(..., weights_only=True)` is compatible with dicts of tensors and scalars — the format used by the voice prompt cache
- `Path.is_relative_to()` is available in Python 3.9+, matching the project's Python 3.12 target (`backend/STYLE_GUIDE.md`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved path validation logic for serving static web application files, ensuring more robust and semantically accurate file access control mechanisms throughout the application platform.
  * Enhanced safety protocols when reconstructing cached voice prompts from persistent disk storage, mitigating potential deserialization vulnerabilities and protecting the system against malformed or corrupted cache data.

### DIFF
--- a/backend/app.py
+++ b/backend/app.py
@@ -135,7 +135,7 @@ def _mount_frontend(application: FastAPI) -> None:
     async def serve_spa(full_path: str):
         file_path = (frontend_dir / full_path).resolve()
         # Guard against path traversal — only serve files inside frontend_dir
-        if full_path and file_path.is_file() and str(file_path).startswith(str(frontend_dir)):
+        if full_path and file_path.is_file() and file_path.is_relative_to(frontend_dir):
             return FileResponse(file_path)
         return FileResponse(frontend_dir / "index.html", media_type="text/html")
 
--- a/backend/utils/cache.py
+++ b/backend/utils/cache.py
@@ -64,7 +64,7 @@ def get_cached_voice_prompt(
     cache_file = _get_cache_dir() / f"{cache_key}.prompt"
     if cache_file.exists():
         try:
-            prompt = torch.load(cache_file)
+            prompt = torch.load(cache_file, weights_only=True)
             _memory_cache[cache_key] = prompt
             return prompt
         except Exception:


afurm:
The weights_only=True fix is correct and aligns with PyTorch 2.6+ best practices. However, the PR description characterizes the cache contents as "basic Python types (strings, ints, floats)" — this is slightly misleading since the .prompt files store dicts containing torch.Tensor objects. The safe unpickler does handle tensors, so this is not a functional issue, but noting that tensor-containing structures are fully supported by weights_only=True would make the rationale more accurate.

afurm:
The is_relative_to() change correctly addresses the prefix-collision case. The existing Path.resolve() already handles ../ traversal attacks. This fix targets the narrower scenario where a path like /app/frontend_evil/ would incorrectly pass the string-prefix check because "/app/frontend_evil".startswith("/app/frontend") is True. The combination of resolve() + is_relative_to() provides sound defense-in-depth.

Issue #542, opened 2026-04-23
Docker - Could not create HuggingFace cache directory


NiCr42:
Hello, 

Thanks for this lovely software !

I'm using the docker compose version and when launching, voicebox tells me that the directory `/home/voicebox/.cache/huggingface/hub` does not have the required permissions. 

Configuration: 
- OpenSUSE Tumbleweed (Linux 6.19.11-1-default (x86_64)) 
- Docker: 29.2.1-ce 
- Docker compose: 5.1.0 
- Voicebox 0.4.5 

Is there a warning to make pre-build docker images available (cuda / rocm) as was the case previously ?

```bash
[+] up 3/3
 ✔ Image voicebox-voicebox       Built                                                       503.0s
 ✔ Network voicebox_voicebox-net Created                                                     2.1s
 ✔ Container voicebox            Recreated                                                   14.8s
Attaching to voicebox
voicebox  | INFO:     Frontend: serving SPA from /app/frontend
voicebox  | INFO:     Started server process [1]
voicebox  | INFO:     Waiting for application startup.
voicebox  | INFO:     Voicebox v0.4.5 starting up
voicebox  | INFO:     Python 3.11.15 on Linux 6.19.11-1-default (x86_64)
voicebox  | INFO:     Database: /app/data/voicebox.db
voicebox  | INFO:     Data directory: /app/data
voicebox  | INFO:     Profiles: 1, Generations: 0
voicebox  | INFO:     Backend: PYTORCH
voicebox  | INFO:     GPU: None (CPU only)
voicebox  | WARNING:     Could not create HuggingFace cache directory: [Errno 13] Permission denied: '/home/voicebox/.cache/huggingface/hub'
voicebox  | INFO:     Ready
voicebox  | INFO:     Application startup complete.
voicebox  | INFO:     Uvicorn running on http://0.0.0.0:17493 (Press CTRL+C to quit)
```

ricardo-ha:
I had  the same problem, you can resolve it with 
``` Dockerfile
# Create HuggingFace cache directory for the named volume
RUN mkdir -p /home/voicebox/.cache/huggingface \
    && chown -R voicebox:voicebox /home/voicebox/.cache/huggingface
```
Before `USER voicebox`

virtuallyunknown:
Getting the same error on startup with the latest version. The suggested fix by @ricardo-ha doesn't work because there is no `USER voicebox` line in the Dockerfile anymore.

Any tips? I am unable to download any models.

[...]

PR #584, opened 2026-04-28
Fix/docker permision

ricardo-ha:
I resolve this Docker problem:
#542 

**Fix permission issues and add GPU support**
The original Dockerfile creates the voicebox user with a system-assigned UID, which causes permission errors in two places:

1. HuggingFace cache — Docker initializes named volumes as root when the target directory doesn't exist in the image, so the voicebox user can't write to it.

2. Bind-mounted output directory — The unpredictable UID doesn't match the host user, causing soundfile to fail when saving generated audio.

**Changes:**

- Pin voicebox to UID/GID 1000 so bind-mounts work without manual chown on the host.

- Pre-create /home/voicebox/.cache/huggingface with correct ownership so the named volume inherits the right permissions.

## Summary by CodeRabbit

* **Chores**
  * Enhanced Docker container user setup with deterministic ID assignment for improved consistency
  * Initialized HuggingFace model cache directory in container environment
  * Added optional NVIDIA GPU support configuration to deployment settings (currently disabled)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -45,9 +45,8 @@ RUN pip install --no-cache-dir --prefix=/install \
 FROM python:3.11-slim
 
 # Create non-root user for security
-RUN groupadd -r voicebox && \
-    useradd -r -g voicebox -m -s /bin/bash voicebox
-
+RUN groupadd -r -g 1000 voicebox && \
+    useradd -r -g voicebox -u 1000 -m -s /bin/bash voicebox
 WORKDIR /app
 
 # Install only runtime system dependencies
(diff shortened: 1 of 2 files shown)


Xarianne:
I hit the same issue myself when building on my machine. But can I suggest a change to this PR?

The pre-created cache directory approach works, but pinning the UID to 1000 is fragile — it assumes the host user is always UID 1000, which isn't guaranteed (e.g. systems where the primary user is UID 1001+, or where Docker runs with a different user namespace).

A more portable approach would be to `chown` the data and cache directories in the entrypoint script before the `gosu` privilege drop. The entrypoint runs as root, so it can fix ownership regardless of what UID the host user has:

```sh
chown -R voicebox:voicebox /app/data
chown -R voicebox:voicebox /home/voicebox/.cache
```
This works for me and I believe is a standard Docker pattern for named volumes and bind mounts.

This would also address the problem in PR #677

PR #602, opened 2026-05-02
fix(transcription): re-encode uploaded audio as PCM WAV before Whisper

### LINKED ISSUE #371: Transcription fails every time
Transcription fails every time. Not sure if this is a bug or user error on my part somehow (I am not technically savvy). I have Whisper Base and Whisper Turbo downloaded, and am using Chatterbox Turbo. This is on an M1 Mac. The error reads, "Transcription failed. Could not open/decode file."

ETA: This is an awesome program, though, very much appreciate you created it, it works very well.

kashmo:
Recorded audio from macOS/WKWebView arrives as mp4/AAC (Safari's MediaRecorder fallback) but was passed directly to mlx-audio Whisper with a .wav extension, causing miniaudio to throw:

  Error -3 while decompressing data: incorrect header check

Fix: after saving the raw upload, decode it with librosa (which handles any container format) and re-write a proper PCM WAV to a second temp file. That clean WAV is what gets passed to Whisper. Both temp files are deleted in the finally block.

Closes #371

## Summary by CodeRabbit

* **Bug Fixes**
  * Transcription now re-encodes uploaded audio to a consistent PCM WAV format before processing and ensures all temporary files are removed, improving reliability across varied audio types.
* **Documentation**
  * Added a step-by-step macOS (Apple Silicon) setup guide for running the local transcription backend, with prerequisites, install/startup commands, verification checks, and troubleshooting notes.

### DIFF
--- a/setup-for-claude.md
+++ b/setup-for-claude.md
@@ -12,6 +12,20 @@ You are helping a user set up a fixed version of the Voicebox AI voice studio on
 
 ---
 
+## Step 0 — Install Official Voicebox App
+
+The fixed backend plugs into the official Voicebox desktop app — the app itself is not built from source here. The user must have it installed first.
+
+Ask the user: **"Do you have Voicebox.app installed in your Applications folder?"**
+
+If not, have them:
+1. Download the DMG from **https://voicebox.sh/download/mac-arm** (Apple Silicon) or **https://voicebox.sh/download/mac-intel** (Intel Mac)
+2. Open the DMG and drag Voicebox to Applications
+3. Launch it once and let it finish its first-run setup
+4. Close it again before continuing — we need to start our backend before the app opens
+
+---
+
 ## Step 1 — Check Prerequisites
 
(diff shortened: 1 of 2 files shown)


setup-for-claude.md
jamiepine, line 1: This file should not be committed as it includes instructions to clone a fork

jamiepine:
The bug you diagnosed here is real — uploaded audio hitting Whisper in the wrong container was behind the whole 30s-cutoff cluster (#609/#626) — and it's now fixed on main via #957, which transcodes at the route boundary. So the transcription issue itself is resolved; thanks for the early diagnosis.

Separately: this PR bundles a `setup-for-claude.md` that instructs AI coding agents to download the DMG and point the installed app at a fork's backend. Regardless of intent, we can't carry docs in-repo that direct users' agents to third-party backends — that's a trust boundary we have to hold, especially given this project's history with impersonation. Please keep any future PRs scoped to the code change itself.

PR #622, opened 2026-05-07
fix(linux-audio): select monitor device by name instead of PULSE_SOURCE env var

### LINKED ISSUE #471: Linux audio: replace std::env::set_var(PULSE_SOURCE) with thread-safe alternative
Follow-up to #457.

The Linux audio capture path sets `PULSE_SOURCE` via `std::env::set_var` inside a spawned capture thread:

https://github.com/jamiepine/voicebox/blob/main/tauri/src-tauri/src/audio_capture/linux.rs

### Problem

- `std::env::set_var` is **not safe** in multi-threaded programs on Unix. Rust 2024 Edition marks it as `unsafe fn` for this reason. The underlying libc `setenv` / `environ` are not thread-safe, so calling it while other threads (tokio runtime, webview, Tauri plugins) may read the environment is a data race / UB risk.
- `PULSE_SOURCE` is never unset, so any future cpal/ALSA init in the same process silently routes to the (possibly stale) monitor. If the user switches default output mid-session, capture keeps targeting the old monitor.

### Options

1. **Pass the source name directly to cpal** if the PulseAudio backend exposes a device-by-name API (preferred).
2. **Shell out to `parec` / `pw-record`** with `Command::env("PULSE_SOURCE", source_name)` so only the child sees the env var; pipe PCM back over stdout.
3. **Set `PULSE_SOURCE` once at app startup** before any threads spawn, with explicit cleanup — least invasive but still leaky.

### Priority

Medium — current behaviour works in practice (single capture at a time, tests clean on PipeWire) and fixes the prior "monitor falls back to microphone" regression. But it's latent UB under the current threading model and will trip any future static-analysis/sanitizer runs.

mvanhorn:
## What changed

`tauri/src-tauri/src/audio_capture/linux.rs`: drop the `std::env::set_var("PULSE_SOURCE", ...)` call inside the spawned capture thread and replace it with a direct device lookup against `host.input_devices()`. The pactl-derived monitor name is matched against cpal-published device names with a four-tier cascade: exact match, suffix match, fuzzy "monitor" name match, then `default_input_device()` as the last resort.

Closes #471

## Why

The issue body lays out the root cause: `std::env::set_var` is unsafe in multithreaded programs. The underlying libc `setenv`/`environ` are not thread-safe, and Rust 2024 has formalised this by marking it `unsafe fn`. Calling it inside a thread that runs concurrently with the tokio runtime, the webview, and Tauri plugins is a data race on `environ`. It also leaks process state — `PULSE_SOURCE` is never unset, so any later cpal/ALSA init in the same process silently routes through the stale monitor.

Option 1 from the issue (pass the source name directly to cpal) is what this PR does. cpal does not have a "device by exact name" API, but `host.input_devices()` enumerates the same PulseAudio sources that `pactl list short sources` reports, so an in-process name match achieves the same routing without touching the environment.

The selection logic prefers an exact pactl source-name match (covers PulseAudio/PipeWire on stock Linux), falls back to a suffix match (covers cpal naming variations), then to the original fuzzy "monitor" search (preserves the prior `pactl`-unavailable code path verbatim), and finally to `default_input_device()` so the failure mode is unchanged.

## Verification

`cargo fmt --check` is clean for this file (the existing `tests/audio_capture_test.rs` formatting issue is pre-existing on `main` and not touched here). The module is gated `#[cfg(target_os = "linux")]`, so local `cargo check` on macOS skips compilation; CI on Linux should validate the type-check.

## Bug Fixes
* Improved Linux audio capture reliability by enhancing monitor-source detection with a prioritized, name-based selection strategy (including exact/suffix and monitor-pattern matching).
* Updated capture start behavior to automatically route from the appropriate Pulse/PipeWire default source to the chosen monitor source when necessary, then restore the previous default after capture begins/ends.
* Adjusted support checks so capture is enabled when the correct monitor source and an applicable input device are present.


mvanhorn:
CodeRabbit's concern was valid: CPAL's Linux host is ALSA, so matching pactl source names against ALSA PCM names mostly can't work (and cpal has no PulseAudio/PipeWire host to switch to, so the literal suggestion isn't available). Reworked in dca6c38:

- Direct monitor matching stays first: exact, suffix, and normalized pactl-name tiers plus an ALSA monitor-alias tier.
- When only the ALSA `pulse` PCM is viable, the PulseAudio/PipeWire default source is pointed at the discovered monitor just for the stream-open window (the pulse PCM binds its source at open), then restored immediately after start; a Drop guard restores it on every error path.
- If no monitor path exists, capture now fails with a descriptive error instead of silently recording the microphone. That last part is a deliberate behavior change; happy to soften it if you'd rather keep the old fallback.

Selection tier is logged so bug reports show which path fired. Linux-only code, so verified by review and rustfmt here; CI will compile-check.

mvanhorn:
Addressed in 5399b30 - the bare contains("monitor") arm is gone, keeping only the precise alias patterns.

PR #629, opened 2026-05-09
fix(refinement): preserve source language in transcript refinement

### LINKED ISSUE #620: 报告bug：自动精修转录，每次生成的必定是英文：Bug Report: Auto-Refinement Transcription Always Generates English
报告bug：自动精修转录，每次生成的必定是英文
报告bug2：连续提交语音样本时，每次点击提交，上次的选择的文件还在选择文件框里
提交建议1：希望增加批量提交语音样本选项
提交建议2：希望语音转录能自行修改错误文本

English translation:

Bug Report: The auto-refinement transcription always generates English text every time.
Bug Report 2: When submitting voice samples consecutively, each time you click submit, the file selected last time still remains in the file selection box.
Suggestion 1: Please add an option for batch submission of voice samples.
Suggestion 2: I hope the voice transcription allows manual correction of incorrectly transcribed text.

JustInCache:
Fixes #620.

## Root cause

Three compounding issues caused the refinement LLM (Qwen3 0.6B/1.7B/4B) to always output English regardless of the transcript's source language:

1. **No language rule in `_BASE_INSTRUCTIONS`** — the system prompt had no instruction to preserve the original language, so the model defaulted to English (its dominant training language).
2. **English-only few-shot examples** — all 7 entries in `REFINEMENT_EXAMPLES` are English, which primes small models toward English output even when the system prompt says otherwise.
3. **`language` field silently discarded** — `refine_capture` never passed the capture's stored `language` (set by Whisper or the user) to `refine_transcript`, so the known language was dropped before the LLM ever saw it.

## Changes

**`backend/services/refinement.py`**

- Added `- Do not translate...` to the Forbidden section of `_BASE_INSTRUCTIONS` as a universal safeguard that applies even when the language is auto-detected and not explicitly known.
- Added a Mandarin Chinese few-shot example to `REFINEMENT_EXAMPLES` (filler removal + punctuation, matching the style of the English demos) to ground the language-preservation rule for small models that need a concrete demonstration to generalize across languages.
- Added an optional `language: str | None = None` parameter to `refine_transcript`. When provided, a language-lock header is prepended at the very top of the assembled system prompt — before all other instructions — giving it maximum positional weight and overriding the English bias from the few-shot examples.

**`backend/services/captures.py`**

- Passed `row.language` from `refine_capture` into the new `language` parameter of `refine_transcript`.

## Testing

To reproduce the bug and verify the fix: set the capture language to `zh` (or any non-English language), dictate a phrase in that language, then trigger auto-refinement. Before this fix the refined output would be English; after, it stays in the original language.

## Summary by CodeRabbit

* **Bug Fixes**
  * Transcript refinement now preserves the original language of captured transcripts, explicitly preventing translation or language conversion during the refinement process.
  * Enhanced support for non-English transcripts with additional language-specific refinement examples to improve accuracy across diverse languages.

[Review Change Stack](https://app.coderabbit.ai/change-stack/jamiepine/voicebox/pull/629)

### DIFF
--- a/backend/services/captures.py
+++ b/backend/services/captures.py
@@ -195,6 +195,7 @@ async def refine_capture(
         row.transcript_raw or "",
         flags,
         model_size=model_size,
+        language=row.language,
     )
 
     row.transcript_refined = refined
--- a/backend/services/refinement.py
+++ b/backend/services/refinement.py
@@ -155,7 +155,8 @@ Forbidden:
 - Do not summarize, shorten, or omit ideas the speaker expressed.
 - Do not add words, examples, explanations, code, or details the speaker did not say.
 - Do not rephrase or substitute synonyms for the speaker's word choices. Keep their vocabulary.
-- Do not wrap the output in quotes, code fences, or a preamble like "Here is the cleaned version". Output only the cleaned transcript itself."""
+- Do not wrap the output in quotes, code fences, or a preamble like "Here is the cleaned version". Output only the cleaned transcript itself.
+- Do not translate. If the transcript is in Chinese, French, Japanese, Russian, or any other language, your output must be in that same language. Never switch to English or any other language not present in the input."""
 
 _SMART_CLEANUP = """Remove disfluencies and empty filler words that interrupt the flow:
 - Disfluencies: "um", "uh", "er", "hmm", "ah"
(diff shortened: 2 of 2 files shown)


JustInCache:
@jamiepine Please let me know if this looks fine, let me know your feedback.

PR #632, opened 2026-05-11
feat: docker improvements - simplified GPU build, cache persistence, fastmcp

ishan-parihar:
## Summary

Docker and infrastructure improvements across multiple areas:

### 1. Simplified GPU Docker build
`Dockerfile.cuda` has been simplified and now pre-warms TTS engines (kokoro, misaki, etc.) during build time for faster container startup.

### 2. Numba cache persistence
Added a `numba-cache` named volume in `docker-compose.gpu.yml` so Numba JIT compilation results persist across container restarts, avoiding recompilation every launch.

### 3. Cache permission fix
Fixed root-owned cache directories in `docker-entrypoint.sh` that caused startup failures when using named volumes.

### 4. Other improvements
- `.dockerignore` and `Dockerfile` cleanup
- `backend/requirements.txt` updates
- TypeScript model type expansion

## Changes
- `Dockerfile.cuda` - Simplified build, added TTS engine pre-warming
- `docker-compose.gpu.yml` - Added numba-cache volume
- `docker-entrypoint.sh` - Cache permission fix
- `Dockerfile` | `app/src/lib/api/types.ts` | `app/src/lib/utils/format.ts` - Supporting changes

## Release Notes

* **New Features**
  * Added GPU auto-offload monitoring for idle TTS/STT models
  * New external TTS generation endpoint (`/tts/generate`)

* **Improvements**
  * Default TTS model size optimized to 0.6B (previously 1.7B)
  * Docker now supports GPU-based deployment via CUDA

* **Removed**
  * Chatterbox and TADA TTS engines removed
  * Personality field removed from voice profiles

[Review Change Stack](https://app.coderabbit.ai/change-stack/jamiepine/voicebox/pull/632)


Xarianne:
Hi, I think this PR won't work the way it is written. It introduces breaking changes to anything that is not Nvidia.

Switching the base `Dockerfile` from `python:3.11-slim` to `nvidia/cuda:12.8.1-runtime-ubuntu22.04` would break the default CPU build path (`docker compose up`), forcing every user to pull a CUDA base image regardless of whether they have a GPU. No CPU-only Dockerfile is created, the existing one is repurposed to CUDA.

It would also break the ROCm build path. The AMD GPU setup uses the same `Dockerfile` with `PYTORCH_VARIANT=rocm` and the `docker-compose.rocm.yml` overlay. The CPU and ROCm variants share the same base image and build stages; switching the base to CUDA removes that shared foundation. Additionally, `docker-compose.rocm.yml` doesn't exist in this PR's branch, and the modified `Dockerfile` no longer has the `PYTORCH_VARIANT` build-arg mechanism, so even if the overlay were present, its build args would be silently ignored.

I am not the developer, but I think this should be resubmitted, and if so it would be easier to review if the changes were split into focused PRs, e.g. one for the cache permission fix, one for the CUDA Dockerfile, one for the GPU monitor, etc. The permission fix is a few lines and could be merged quickly on its own; the larger changes need more consideration.

PR #656, opened 2026-05-14
feat: add OpenAI-compatible /v1/audio/speech and /v1/models endpoints

neuron-tech-ai:
Adds two endpoints that mirror the OpenAI TTS API:

- `POST /v1/audio/speech` — accepts the same request shape as OpenAI's TTS endpoint (`model`, `input`, `voice`, `response_format`, `speed`)
- `GET /v1/models` — returns available Voicebox models in OpenAI's model list format

**Why this matters:** any application already integrated with OpenAI TTS can point at a local Voicebox instance by changing only the base URL and API key — no code changes on the client side. This makes Voicebox a drop-in local replacement for OpenAI TTS in existing workflows, scripts, and tools.

## New Features
* Added OpenAI-compatible text-to-speech API endpoints for generating audio from text with support for multiple voices and customizable speech parameters.
* Added endpoint to retrieve available text-to-speech models.

[Review Change Stack](https://app.coderabbit.ai/change-stack/jamiepine/voicebox/pull/656)

### DIFF
--- a/backend/routes/__init__.py
+++ b/backend/routes/__init__.py
@@ -23,6 +23,7 @@ def register_routers(app: FastAPI) -> None:
     from .speak import router as speak_router
     from .mcp_bindings import router as mcp_bindings_router
     from .events import router as events_router
+    from .openai_compat import router as openai_compat_router
 
     app.include_router(health_router)
     app.include_router(profiles_router)
@@ -42,3 +43,4 @@ def register_routers(app: FastAPI) -> None:
     app.include_router(speak_router)
     app.include_router(mcp_bindings_router)
     app.include_router(events_router)
+    app.include_router(openai_compat_router)
--- a/backend/routes/openai_compat.py
+++ b/backend/routes/openai_compat.py
@@ -16,9 +16,9 @@ GET /v1/models
 
 Model mapping
 -------------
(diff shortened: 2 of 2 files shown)


neuron-tech-ai:
Addressed all five CodeRabbit review comments:

1. **Model ID typo**: Fixed `gts-4o-mini-tts` → `gpt-4o-mini-tts` in both the module docstring and the `_MODEL_MAP` dict.

2. **`response_format` silently ignored**: Added documentation in the module docstring noting that WAV is always returned and other formats are not yet supported. Raising a 400 for unsupported formats would break existing callers that pass the default; the limitation is now explicitly documented.

3. **`speed` parameter not forwarded**: Added documentation in the module docstring noting that `speed` is accepted for schema compatibility but not yet forwarded to backends. A follow-up implementation PR can wire this through once backend support is added.

4. **Bare `except Exception: pass`**: Replaced with `except Exception as e` that logs a warning (with voice name and exception message) before falling through to the built-in voice.

5. **Wrong fallback `voice_prompt` dict keys**: Fixed the fallback return to use engine-specific keys: Kokoro gets `{"kokoro_voice": kokoro_voice}`, all others (qwen, qwen_custom_voice) get `{"preset_voice_id": kokoro_voice}`.

PR #657, opened 2026-05-14
feat: add platform gating to ModelConfig and UI

neuron-tech-ai:
Some models only run on specific platforms (e.g., MLX models require macOS with Apple Silicon; some CUDA-only models won't run on macOS). Currently these models show up in the UI on all platforms and fail at download or load time with unhelpful errors.

This adds a `platforms` field to `ModelConfig` so each model can declare which platforms it supports. The UI uses this to hide or disable models that won't run on the current platform — users only see models that will actually work for them.

Reduces confusion and support burden from "why won't this model load" issues that are really just platform incompatibility.

## Summary by CodeRabbit

* **New Features**
  * Hardware platform compatibility checking for models and engines with visual lock icons indicating incompatibility
  * Required hardware information now displayed via tooltips for incompatible options
  * Automatic detection of available compute platforms on your system (CUDA, ROCm, MPS, XPU, CPU)
  * Incompatible engines are now disabled and marked as unavailable based on detected platform support

[Review Change Stack](https://app.coderabbit.ai/change-stack/jamiepine/voicebox/pull/657)

### DIFF
--- a/app/src/components/Generation/EngineModelSelector.tsx
+++ b/app/src/components/Generation/EngineModelSelector.tsx
@@ -43,6 +43,22 @@ const ENGINE_DESCRIPTIONS: Record<string, string> = {
   kokoro: '82M params, CPU realtime, 8 langs',
 };
 
+/** Map from engine name to the model_name used by the backend status API. */
+const ENGINE_TO_MODEL_NAME: Partial<Record<string, string>> = {
+  kokoro: 'kokoro',
+  luxtts: 'luxtts',
+  chatterbox: 'chatterbox-tts',
+  chatterbox_turbo: 'chatterbox-turbo',
+};
+
+/** Derive the backend model_name string for an ENGINE_OPTIONS entry. */
+function deriveModelName(opt: (typeof ENGINE_OPTIONS)[number]): string {
+  if (opt.value.includes(':')) {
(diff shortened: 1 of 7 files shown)


neuron-tech-ai:
Addressed all four CodeRabbit review comments:

1. **Nested ternary refactor**: Extracted the `optModelName` logic into a `deriveModelName(opt)` helper function with a `ENGINE_TO_MODEL_NAME` lookup map. The loop body now calls `deriveModelName(opt)` instead of the 10-line nested ternary.

2. **Ruff E402 import ordering**: Moved `from ..utils.platform_detect import get_backend_type, get_supported_platforms` into the top import block (immediately after the `hf_offline_patch` import), before the constant definitions.

3. **Per-variant platform requirements**: Updated `is_engine_platform_compatible` to check across all variants for an engine (returns True if any variant's requirements are met). Updated `load_engine_model` to select the config matching `model_size` before reading `requires`, falling back to `configs[0]` only if no match is found.

4. **Remove unnecessary `intel_extension_for_pytorch` import**: Removed the try/except IPEX import. Native `torch.xpu.is_available()` (PyTorch 2.7+) is used directly with a `hasattr(torch, 'xpu')` guard.

PR #659, opened 2026-05-14
fix: replace deprecated datetime.utcnow() with datetime.now(UTC) throughout

neuron-tech-ai:
`datetime.utcnow()` is deprecated in Python 3.12 and raises `DeprecationWarning`. It also returns a naive datetime (no timezone info), which is a footgun when comparing with timezone-aware datetimes.

Replaces all usages across the codebase with `datetime.now(UTC)`, which returns a timezone-aware UTC datetime and is the correct approach going forward. Pure mechanical replacement — no behavior change.

## Release Notes

**Bug Fixes**
- Updated all system timestamps to use timezone-aware UTC format, improving timestamp consistency and accuracy across profile management, story editing, task tracking, and data import operations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/jamiepine/voicebox/pull/659)

### DIFF
--- a/backend/database/models.py
+++ b/backend/database/models.py
@@ -1,6 +1,6 @@
 """ORM model definitions for the voicebox SQLite database."""
 
-from datetime import datetime
+from datetime import UTC, datetime
 import uuid
 
 from sqlalchemy import Column, String, Integer, Float, DateTime, Text, ForeignKey, Boolean, JSON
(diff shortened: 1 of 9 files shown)


neuron-tech-ai:
Addressed the CodeRabbit review comment:

**Missing `UTC` import causing `NameError`**: The function-scope import at the top of `import_generation_from_zip` was `from datetime import datetime`. Changed to `from datetime import UTC, datetime` so `datetime.now(UTC)` on line 435 resolves correctly at runtime.

PR #660, opened 2026-05-14
fix: MediaRecorder crash on rapid re-record + add sample sort_order and reorder API

neuron-tech-ai:
**Bug fix:** Rapidly stopping and restarting MediaRecorder (e.g., recording a second sample immediately after stopping the first) throws an "already recording" `InvalidStateError`. Added proper state checks before start/stop calls so rapid re-record works reliably.

**Feature addition:** Adds `sort_order` to profile samples and a `PUT /profiles/{id}/samples/reorder` endpoint. Users expect to be able to drag-and-drop reorder their voice samples; without server-side order persistence, the frontend can't save that state.

## Release Notes

* **New Features**
  * Added sample reordering functionality for organizing audio recordings.

* **Documentation**
  * Updated audio sample guidance across all supported languages, clarifying that adding multiple diverse recordings improves voice quality through automatic merging.

* **Improvements**
  * Enhanced error handling with clearer messages for file uploads and validation errors.
  * Database performance optimizations for improved responsiveness and faster data operations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/jamiepine/voicebox/pull/660)


neuron-tech-ai:
Fixed two bugs found during review:

1. **Missing `SampleReorderRequest` model**: The new `PATCH /profiles/{profile_id}/samples/reorder` endpoint referenced `models.SampleReorderRequest`, but this class was never added to `backend/models.py`. This would cause an `AttributeError` at import time, crashing the server on startup. Added `SampleReorderRequest` with a `sample_ids: list[str]` field.

2. **Missing `sort_order` field in `ProfileSampleResponse`**: The DB model gained a `sort_order` column and the TypeScript client types were updated, but the Python `ProfileSampleResponse` Pydantic model was never updated to include it. Serializing samples would either omit the field or raise a validation error depending on Pydantic mode. Added `sort_order: int = 0` to `ProfileSampleResponse`.

Fixed in commit bf0ad3e.

PR #669, opened 2026-05-15
fix: the justfile syntax errror, GPU information cannot be read

xorss:
The absence of `\` prevents the GPU information cannot be read.(Windows)
i don't know JUST. you can ignore this issue if it can work in other development environment. ;D


xorss:
I didn't notice the branch and pushed some code. But it has been withdrawn.

PR #677, opened 2026-05-18
Fix HF cache permissions and add startup fallback

enveryasar:
## Summary by CodeRabbit

* **Chores**
  * Improved Hugging Face model cache initialization with automatic directory setup during startup
  * Added graceful fallback handling if cache directory creation fails
  * Updated container environment configuration for cache directory paths

[Review Change Stack](https://app.coderabbit.ai/change-stack/jamiepine/voicebox/pull/677?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/backend/app.py
+++ b/backend/app.py
@@ -282,6 +282,12 @@ async def _run_startup(application: FastAPI) -> None:
     except Exception as e:
         logger.warning("Could not initialize progress manager event loop: %s", e)
 
+    fallback_cache_dir = config.get_cache_dir() / "huggingface" / "hub"
+    if not os.environ.get("HF_HUB_CACHE"):
+        # Default to /app/data/cache in containers (and data/cache in dev)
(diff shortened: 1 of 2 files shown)


Xarianne:
I hit the same issue and suggested a fix in PR #584 — `chown` the data and cache directories in the entrypoint before the `gosu` privilege drop.

The issue with this PR's approach is that moving the cache to `/app/data` doesn't actually fix the root-ownership problem — `/app/data` is a named volume that's also mounted as `root:root` on first use, which is the same thing causing the SQLite `attempt to write a readonly database` error. The fallback in `app.py` would hit the same permission denied when trying to create `/app/data/cache/huggingface/hub`.

Fixing the ownership in the entrypoint before the privilege drop would resolve both the cache permission issue and the SQLite readonly error without needing the application-level fallback or moving the cache location.

PR #748, opened 2026-06-13
fix(linux): vendor-patch tao 0.34.8 to avoid Wayland startup panic

Kawa99:
## Summary
- On Fedora 44 / GNOME 50 (Wayland-only), `tao` 0.34.8 crashes on startup with `Option::unwrap()` on `None` at `event_loop.rs:448` inside the `CursorIgnoreEvents` handler
- `window.window()` returns `None` on Wayland (no X11-backed GdkWindow), and `input_shape_combine_region` is X11-only — the unwrap panics through glib's main-context channel
- Vendored tao 0.34.8 at `tauri/vendor/tao`, changed `unwrap()` → `if let Some(...)` to silently skip the X11 shape call on Wayland
- Switched `[patch.crates-io]` from `tao-v0.35.3` (git tag, semver-incompatible with `tauri-runtime-wry`'s `"0.34.5"` constraint) to `path = "../vendor/tao"`, ensuring a single tao in the build

## Summary by CodeRabbit

* **Chores**
  * Updated the windowing and event-loop library to a locally maintained version.


J-eremy:
Hitting this same issue on CachyOS (Arch) with an RTX 5080. The `GDK_BACKEND=x11` workaround gets past the Wayland crash, but there's also a gray/blank screen problem that needs `WEBKIT_DISABLE_DMABUF_RENDERER=1` on top of it. Seems like the WebKitGTK DMA-BUF path is broken on NVIDIA. Opened #1029 for the rendering side of things. This PR fixes the unwrap panic which is the right call — the DMA-BUF issue is separate.

PR #808, opened 2026-06-28
fix: harden API/Tauri security and fix MCP tool naming

### LINKED ISSUE #778: Security hardening: DNS-rebinding (Host-header) exposure on the local API/MCP server + over-broad Tauri capabilities — fixes available
Hi — while evaluating Voicebox for a local integration I ran a security pass over the server and the Tauri shell and found two issues worth hardening. I have fixes applied locally against `b35b909` and I'm happy to open a PR.

### 1. (High) No Host-header / DNS-rebinding protection on the local server

`backend/app.py` (`create_app()`) and the mounted MCP ASGI sub-app (`backend/mcp_server/server.py`) listen on `127.0.0.1:17493` but don't validate the `Host` header. The unauthenticated REST endpoints (`/speak`, `/transcribe`, `/shutdown`) and the MCP transport are therefore reachable via **DNS rebinding**: a malicious page the user visits can rebind its own hostname to `127.0.0.1` and drive the server from the browser — speak, enumerate/read transcripts, shut it down, and (via `audio_path`) read local files — because the request is loopback-sourced and nothing checks `Host`.

**Fix** — add Starlette's `TrustedHostMiddleware` with a loopback allow-list, on both the outer FastAPI app and the FastMCP sub-app (`http_app(...)` accepts a `middleware=` list):

```python
from starlette.middleware import Middleware
from starlette.middleware.trustedhost import TrustedHostMiddleware

ALLOWED = ["127.0.0.1", "localhost", "127.0.0.1:17493", "localhost:17493"]
application.add_middleware(TrustedHostMiddleware, allowed_hosts=ALLOWED)
mcp_app = mcp.http_app(
    path="/", transport="http",
    middleware=[Middleware(TrustedHostMiddleware, allowed_hosts=ALLOWED)],
)
```

After this, `Host: evil.com` returns `400` while `Host: 127.0.0.1:17493` is served normally (I verified this with an in-process Starlette `TestClient`).

*(Optional defense-in-depth: gate the `audio_path` file read on a per-session token rather than a bare loopback check. With the Host gate in place this is no longer the primary exposure — it closes the residual case of a non-browser client that can set an arbitrary `Host`.)*

### 2. (High) Over-broad Tauri capabilities

`tauri/src-tauri/capabilities/default.json` + `tauri.conf.json` grant the webview more than the app's own Rust invokes:

- `shell:allow-execute` + `shell:allow-spawn` — arbitrary process execution, not used by the app
- `fs:read-all` + `fs:write-all` — whole-filesystem access
- `shell.open` scope `".*"` — open any URL/scheme
- `"csp": null` — no Content-Security-Policy
- `remote.urls: ["http://localhost:*"]` — any localhost port

A compromised or hostile web context in the webview could leverage these.

**Recommended fix** — drop `shell:allow-execute`/`shell:allow-spawn`; **narrow `shell.open` to only the `x-apple.systempreferences:` schemes the app actually opens** (avoid a catch-all or a broad absolute-path scope — a permissive `shell.open` is itself a vector); replace `fs:read-all`/`fs:write-all` with scoped fs permissions for the app-data dir (keep a *scoped* read if the app loads voice/audio files at runtime); set a real CSP (`default-src 'self'; connect-src http://127.0.0.1:17493 …`); and pin `remote.urls` to the specific `127.0.0.1:17493` origin (+ the Vite dev port).

*Note:* I applied these locally but haven't exercised a full signed Tauri build (the build pulls a prebuilt `voicebox-server` sidecar I don't have), so the capability changes are static-reviewed — worth a build-test on your side before merging.

### Offer

Happy to open a PR with the changes, or you're welcome to take them directly — whichever is easier for you.

### LINKED ISSUE #790: MCP tool names contain "." which violates Claude Desktop's name pattern ^[a-zA-Z0-9_-]{1,64}$
## Summary

The Voicebox MCP server exposes tool names that use a dot (`.`) as a namespace separator (e.g. `voicebox.speak`). Claude Desktop rejects these because tool names must match the pattern `^[a-zA-Z0-9_-]{1,64}$`, which does not allow `.`. As a result, connecting to the Voicebox MCP server breaks the entire tool list with the following error:

tools.87.FrontendRemoteMcpToolDefinition.name: String should match pattern '^[a-zA-Z0-9_-]{1,64}$'

## Affected tool names

All four exposed tools are affected:

- `voicebox.speak`
- `voicebox.transcribe`
- `voicebox.list_captures`
- `voicebox.list_profiles`

## Steps to reproduce

1. Run Voicebox v0.5.0 (CUDA backend) with the MCP server mounted at `/mcp`.
2. Add it as a remote MCP server in Claude Desktop.
3. Claude Desktop fetches the tool list and rejects it because the names contain `.`.
4. With Voicebox enabled, the whole tool list fails validation; disabling Voicebox makes the error disappear.

## Expected behavior

Tool names should match `^[a-zA-Z0-9_-]{1,64}$`. Using an underscore (or hyphen) as the namespace separator would fix this, e.g. `voicebox_speak`, `voicebox_transcribe`, `voicebox_list_captures`, `voicebox_list_profiles`.

## Environment

- Voicebox v0.5.0 (PyTorch / CUDA backend)
- OS: Windows 11 (AMD64)
- Python 3.12.10
- GPU: NVIDIA GeForce RTX 3060
- MCP transport: StreamableHTTP, mounted at `/mcp`
- Client: Claude Desktop

## Notes

Verified the tool names directly via a `tools/list` call against the MCP endpoint; all returned names use the `voicebox.<tool>` dot format. The MCP spec / Claude's validator restricts tool names to letters, digits, underscore and hyphen, so the dot is the root cause.

jashwanth-reddy-g:
Fixes #778 and Fixes #790

## Summary by CodeRabbit

* **Security**
  * Tightened host, network, and app security settings to better restrict allowed connections and browser/app resource access.
  * Limited opening external links to a specific system settings scheme.
* **Changes**
  * Updated tool names used by the app’s voice features.
  * Reduced default app permissions in the desktop build, removing some broad system and file access.

### DIFF
--- a/backend/app.py
+++ b/backend/app.py
@@ -71,12 +71,21 @@ def create_app() -> FastAPI:
     """Create and configure the FastAPI application."""
     from .mcp_server.server import build_mcp_server, compose_lifespan
     from .mcp_server.context import ClientIdMiddleware
+    from starlette.middleware import Middleware
+    from starlette.middleware.trustedhost import TrustedHostMiddleware
 
     # Build the MCP app up-front so we can wire its lifespan into FastAPI's —
     # FastMCP's Streamable HTTP transport only works if its session manager
     # runs inside the parent ASGI lifespan.
     mcp = build_mcp_server()
-    mcp_app = mcp.http_app(path="/", transport="http")
+    mcp_app = mcp.http_app(
+        path="/", 
+        transport="http",
+        middleware=[Middleware(
+            TrustedHostMiddleware, 
+            allowed_hosts=["127.0.0.1", "localhost", "127.0.0.1:17493", "localhost:17493"]
+        )],
+    )
 
     @asynccontextmanager
     async def voicebox_lifespan(app: FastAPI):
@@ -106,6 +115,10 @@ def create_app() -> FastAPI:
 
     _configure_cors(application)
     application.add_middleware(ClientIdMiddleware)
+    application.add_middleware(
+        TrustedHostMiddleware,
+        allowed_hosts=["127.0.0.1", "localhost", "127.0.0.1:17493", "localhost:17493"]
+    )
     register_routers(application)
     application.mount("/mcp", mcp_app)
     logger.info("MCP: mounted at /mcp")
--- a/backend/mcp_server/tools.py
+++ b/backend/mcp_server/tools.py
@@ -36,7 +36,7 @@ def register_tools(mcp: FastMCP) -> None:
     """Attach all Voicebox tools to the given FastMCP instance."""
 
     @mcp.tool(
-        name="voicebox.speak",
+        name="voicebox_speak",
         description=(
             "Speak text in a Voicebox voice profile. Returns a generation id "
             "the caller can poll at /generate/{id}/status. Audio plays on the "
@@ -105,7 +105,7 @@ def register_tools(mcp: FastMCP) -> None:
             db.close()
 
     @mcp.tool(
-        name="voicebox.transcribe",
+        name="voicebox_transcribe",
         description=(
             "Transcribe an audio clip to text using Voicebox's local Whisper. "
             "Pass exactly one of `audio_base64` (bytes as base64) or "
@@ -163,7 +163,7 @@ def register_tools(mcp: FastMCP) -> None:
             tmp_path.unlink(missing_ok=True)
 
     @mcp.tool(
-        name="voicebox.list_captures",
+        name="voicebox_list_captures",
         description=(
             "List recent voice captures (dictations, recordings, uploads) "
             "with their transcripts. Most-recent first."
@@ -191,7 +191,7 @@ def register_tools(mcp: FastMCP) -> None:
             db.close()
 
     @mcp.tool(
-        name="voicebox.list_profiles",
+        name="voicebox_list_profiles",
         description=(
             "List available voice profiles (both cloned voices and presets). "
             "Use the returned `name` with voicebox.speak(profile=...)."
--- a/tauri/src-tauri/capabilities/default.json
+++ b/tauri/src-tauri/capabilities/default.json
@@ -5,7 +5,7 @@
   "platforms": ["linux", "macOS", "windows"],
   "windows": ["main", "dictate"],
   "remote": {
-    "urls": ["http://localhost:*"]
+    "urls": ["http://127.0.0.1:17493", "http://localhost:5173", "http://127.0.0.1:5173"]
   },
   "permissions": [
     "core:default",
@@ -14,15 +14,11 @@
     "core:webview:default",
     "core:webview:allow-internal-toggle-devtools",
     "shell:allow-open",
-    "shell:allow-execute",
-    "shell:allow-spawn",
     "updater:default",
     "process:default",
     "dialog:default",
     "dialog:allow-save",
     "dialog:allow-open",
-    "fs:default",
-    "fs:read-all",
-    "fs:write-all"
+    "fs:default"
   ]
 }
--- a/tauri/src-tauri/tauri.conf.json
+++ b/tauri/src-tauri/tauri.conf.json
@@ -36,7 +36,7 @@
   "app": {
     "macOSPrivateApi": true,
     "security": {
-      "csp": null,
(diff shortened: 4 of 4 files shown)


timharris707:
Thanks for putting #808 together. I used it as the starting point for a local hardening fork and validated the same general direction there: Host-header protection, tighter Tauri capability/CSP settings, and underscore MCP tool names.

One gap I kept separate is #809 (`/models/migrate` accepting arbitrary destinations). I opened a focused PR for that here: #815. It constrains migration to Voicebox-owned model storage, rejects outside/traversal/symlink/source-contained destinations, and stops delete-before-move overwrite behavior.

In the local fork I also had to replace a couple of UX paths after removing broad Tauri filesystem/shell powers: capture export now goes through a scoped native `save_file` command, and folder opening goes through a scoped `open_path` command instead of general webview `fs` writes / unrestricted open. I did not include those in #815 because they belong with the Tauri permission hardening work rather than the model-migration route fix.

Validation I ran locally for #809:

- focused pytest for migration containment paths
- ruff `I,F` + format over touched files
- bundled server/app smoke proving loopback bind + hostile Host rejection in the hardened fork

Happy to split the Tauri scoped save/open replacements into a follow-up PR if that would help land #808 cleanly.

Krustin:
While waiting for this to merge, here's a client-side workaround for anyone else blocked on this in Claude Desktop: a small Python stdio proxy that sits between Claude Desktop and the bundled `voicebox-mcp` shim, rewriting tool names bidirectionally (`voicebox.speak` ↔ `voicebox_speak`, etc.) on the JSON-RPC stream. Everything else passes through untouched.

```python
#!/usr/bin/env python3
import json, os, re, subprocess, sys, threading

VOICEBOX_MCP_BIN = os.environ.get(
    "VOICEBOX_MCP_BIN",
    os.path.expanduser("~/Applications/Voicebox.app/Contents/MacOS/voicebox-mcp"),
)
DOT_TO_UNDERSCORE = re.compile(r"^voicebox\.")
UNDERSCORE_TO_DOT = re.compile(r"^voicebox_")

def to_claude_name(name):
    return DOT_TO_UNDERSCORE.sub("voicebox_", name, count=1)

def to_voicebox_name(name):
    return UNDERSCORE_TO_DOT.sub("voicebox.", name, count=1)

def main():
    child = subprocess.Popen([VOICEBOX_MCP_BIN], stdin=subprocess.PIPE,
                              stdout=subprocess.PIPE, stderr=sys.stderr,
                              text=True, bufsize=1)
    pending, lock = {}, threading.Lock()

    def client_to_child():
        for line in sys.stdin:
            line = line.rstrip("\n")
            if not line: continue
            try: msg = json.loads(line)
            except json.JSONDecodeError:
                child.stdin.write(line + "\n"); child.stdin.flush(); continue
            method = msg.get("method")
            if method in ("tools/call", "tools/list") and "id" in msg:
                with lock: pending[msg["id"]] = method
            if method == "tools/call":
                params = msg.get("params", {})
                if "name" in params:
                    params["name"] = to_voicebox_name(params["name"])
            child.stdin.write(json.dumps(msg) + "\n"); child.stdin.flush()
        child.stdin.close()

    def child_to_client():
        for line in child.stdout:
            line = line.rstrip("\n")
            if not line: continue
            try: msg = json.loads(line)
            except json.JSONDecodeError:
                sys.stdout.write(line + "\n"); sys.stdout.flush(); continue
            method = None
            if msg.get("id") is not None:
                with lock: method = pending.pop(msg["id"], None)
            if method == "tools/list":
                for tool in msg.get("result", {}).get("tools", []):
                    if "name" in tool:
                        tool["name"] = to_claude_name(tool["name"])
            sys.stdout.write(json.dumps(msg) + "\n"); sys.stdout.flush()

    t1 = threading.Thread(target=client_to_child, daemon=True)
    t2 = threading.Thread(target=child_to_client, daemon=True)
    t1.start(); t2.start()
    child.wait(); t2.join(timeout=2)

if __name__ == "__main__":
    main()
```

Point Claude Desktop's `claude_desktop_config.json` at this script instead of `voicebox-mcp` directly:

```json
"voicebox": {
  "command": "/path/to/python3",
  "args": ["/path/to/voicebox-mcp-bridge.py"],
  "env": {
    "VOICEBOX_CLIENT_ID": "claude-desktop",
    "VOICEBOX_MCP_BIN": "/path/to/Voicebox.app/Contents/MacOS/voicebox-mcp"
  }
}
```

Confirmed working: `tools/list` returns clean underscored names, `tools/call` round-trips correctly (tested against `voicebox_list_profiles` on macOS/MLX). This is obviously just a stopgap — the real fix is the underscore rename in this PR, no need for anyone to actually run a proxy once it merges.

Issue #708, opened 2026-05-26
[BUG] Download progress bars display incorrect values when downloading multiple models concurrently (Windows)


HiccupLY:
## Description

When downloading multiple models simultaneously from the "Models" page in the Windows UI, the progress bars display incorrect values. The progress of a single file's download overwrites the progress values of all other active downloads, causing all progress bars to fluctuate erratically between each other's values rather than reflecting their respective individual progress.

## Steps to Reproduce

1. Open the Voicebox desktop application on Windows.
2. Navigate to the "Models" page.
3. Initiate downloads for two or more models at the same time.
4. Observe the progress bars for each download.

## Expected Behavior

Each model's progress bar should independently reflect its own download progress. When multiple models are downloading concurrently, each progress bar should display only its corresponding file's download percentage.

## Actual Behavior

When a download progress update is received for any one file, that file's progress value appears to be applied to **all** active download progress bars simultaneously. As different files receive progress updates in rapid succession, all visible progress bars jump between each other's values, making it impossible to determine the actual download progress of any individual file.

## Environment

- **OS:** Windows
- **Voicebox Version:** 0.5.0
- **Page:** Models → Download

## Additional Context

This appears to be a state management issue where the download progress callback does not correctly scope its update to the specific model being downloaded, instead broadcasting the progress value to all active download items.

PR #829, opened 2026-07-01
fix(backend): make HF download progress tracking safe under concurrent downloads

### LINKED ISSUE #784: [bug]The progress will be disordered when multiple models are downloaded simultaneously.
As shown in the figure, the progress bars will synchronize inexplicably when multiple models are downloaded simultaneously and be forced to the same progress.
Image

Image

singlaamitesh:
## Summary

`HFProgressTracker.patch_download()` monkey-patches the process-wide `tqdm` class every time it's called, binding the tracker instance into a closure. When two models download at the same time (each running its blocking HF call on its own thread), the second call's `patch_download()` silently overwrites the first call's patched class — both downloads' tqdm updates end up dispatching to whichever tracker patched most recently, so:

- Progress bars for two simultaneous downloads sync up / show each other's numbers (#784).
- Whichever download finishes first runs its `finally` block and tears down the monkey-patch globally, killing progress tracking for the other download that's still in flight.

## Fix

- The tqdm monkey-patch is now installed **at most once per process**, refcounted across concurrent `patch_download()` calls — only the first concurrent download installs it, only the last one to finish tears it down.
- Dispatch now goes through a `threading.local()` "current tracker for this thread" instead of a closure variable, so each thread's tqdm updates always resolve to *its own* tracker. Each download already runs its blocking HF call on a dedicated thread, so this cleanly isolates concurrent downloads without needing per-download class patching.
- Per-tracker state (`_file_sizes`, `_active_tqdms`, etc.) is unchanged — it was already correctly scoped to the tracker instance; the bug was purely in how tqdm dispatched *to* the right instance.

## Test plan

- [x] Added `test_concurrent_downloads_dont_cross_contaminate` to `backend/tests/test_progress.py`: runs two `HFProgressTracker`s on two threads with a `threading.Barrier` to force real overlap, and asserts each tracker only ever sees its own filename/total, both finish at their own file size, and the patch is fully restored once both are done.
- [x] Verified this new test **fails** against the pre-fix code (tracker A observed tracker B's filename) and **passes** against this fix.
- [x] Ran the full `backend/tests/test_progress.py` suite — all tests pass except `test_hf_progress_tracker`, which is a pre-existing failure on `main` unrelated to this change (its dummy file size of 1000 bytes never crosses the tracker's existing 1MB `MIN_TOTAL_BYTES` reporting threshold — confirmed by running it against unmodified `main`). Left it out of scope here to keep this PR focused on the concurrency fix.

Fixes #784

## Summary by CodeRabbit

* **New Features**
  * Progress tracking now works correctly for multiple downloads happening at the same time.

* **Bug Fixes**
  * Fixed progress updates from different downloads overwriting or mixing with each other.
  * Improved reliability when starting and finishing overlapping downloads, including nested progress handling.
  * Added coverage for concurrent download scenarios to prevent regressions.

### DIFF
--- a/backend/tests/test_progress.py
+++ b/backend/tests/test_progress.py
@@ -4,6 +4,7 @@ Test script to debug model download progress tracking.
 
 import asyncio
 import json
+import threading
 import time
 from typing import List, Dict
 import logging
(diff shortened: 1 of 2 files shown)


singlaamitesh:
This also fixes the duplicate report in #708 (same root cause — concurrent downloads sharing a corrupted global tqdm patch).

PR #831, opened 2026-07-01
fix: gate macOS-only keyboard_layout symbols behind cfg

youtsuhodev:
## Description

This PR fixes `dead_code` compiler warnings in `keyboard_layout.rs` by properly gating all macOS-specific symbols behind `#[cfg(target_os = "macos")]`.

### Problem

The following symbols were defined at module level and compiled on all platforms, but only used on macOS:

- `FALLBACK_V_KEYCODE` — the `kVK_ANSI_V` keycode constant
- `V_KEYCODE` — an `AtomicU16` holding the resolved keycode
- `paste_keycode_v()` — the public accessor for the resolved keycode
- `AtomicU16` and `Ordering` imports — only needed by the above

On Windows/Linux, these generated 3 warnings:
warning: constant FALLBACK_V_KEYCODE is never used warning: static V_KEYCODE is never used warning: function paste_keycode_v is never used

### Solution

Moved all four items behind `#[cfg(target_os = "macos")]`. The `mod macos` (already cfg-gated) uses them via `use super::*`, so macOS builds are unaffected.

### Files changed

- `tauri/src-tauri/src/keyboard_layout.rs` — added `#[cfg(target_os = "macos")]` on the import, constant, static, and public function
- `tauri/src-tauri/Cargo.toml`, `tauri/src-tauri/gen/schemas/desktop-schema

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved keyboard input handling so paste-related key detection is only compiled and used on macOS, matching platform behavior more reliably.
  * Reduced the chance of unexpected behavior on non-macOS platforms by avoiding macOS-specific keycode handling there.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -57,7 +57,6 @@
         "react-dom": "^18.3.0",
         "react-hook-form": "^7.53.0",
         "react-i18next": "^17.0.4",
-        "react-qr-code": "^2.0.18",
         "react-sound-visualizer": "^1.4.0",
         "tailwind-merge": "^2.5.4",
         "wavesurfer.js": "^7.0.0",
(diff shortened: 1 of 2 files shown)


KirobotDev:
LGTM Rust this good

Issue #626, opened 2026-05-08
Whisper Transcription Cutoff: Only first paragraph/30 seconds of imported WAV is transcribed


reverendrichie-code:
Description:
When importing an external .wav file for transcription, Voicebox 0.5.0 successfully processes the beginning of the audio but stops prematurely. Regardless of the total file length (tested with files ranging from 1 to 10 minutes), the output contains only the first paragraph or approximately the first 30 seconds of speech.

Steps to Reproduce:

Go to the Story/Timeline view.

Import an external .wav file (longer than 30 seconds).

Select "Transcribe" using any Whisper model (tested with Large, Medium, and Turbo).

Observe the output in the text area.

Expected Behavior:
The entire audio file should be transcribed.

Actual Behavior:
Transcription stops after the first ~30 seconds (roughly one paragraph), leaving the rest of the audio untranscribed.

Environment:

Voicebox Version: 0.5.0

OS: Windows 11

Hardware: NVIDIA RTX 5060 Ti 16GB (Compute Capability 9.0)

Models Tested: Whisper Large, Medium, Turbo

Additional Context:
I suspect the backend may be missing a chunking or sliding-window mechanism for imported files. While the new "Capture" feature works for short bursts, imported audio appears to hit the default 30-second Whisper window limit and then stop, rather than proceed to the next segment.

Fessoid:
Yeap, i have same problem, only first 30 sec are recognizable

djismgaming:
Indeed, I see the same behavior on v0.5.0 while importing a 1 minute audio. The transcription is correct but only the output of the first 30 seconds of the audio file appear in the Captures tab entry.

edmiidz:
Confirming this on **macOS / Apple Silicon** too (from a `cabef1b` source build), so it's not platform-specific — and pinpointing the root cause.

**Root cause:** the STT backends feed the *entire* audio into the processor and call `model.generate()` **once**, with no chunking. Whisper only attends to a single **30s** window, so the processor pads/truncates to 30s and only the first ~30s is ever transcribed.

`backend/backends/pytorch_backend.py` (`PyTorchSTTBackend.transcribe` → `_transcribe_sync`):

```python
inputs = self.processor(audio, sampling_rate=16000, return_tensors="pt")  # pads/truncates to 30s
inputs = inputs.to(self.device)
...
predicted_ids = self.model.generate(inputs["input_features"], **generate_kwargs)  # single 30s window
[...]

PR #832, opened 2026-07-01
feat(stt): transcribe long audio via overlapping Whisper chunks

deepakmaseeh:
## Summary

- Adds `backend/utils/whisper_stt.py` to window audio longer than Whisper'\''s ~30s limit into overlapping 30s chunks (5s stride)
- Wires `PyTorchSTTBackend.transcribe()` through the shared helper so Captures, `/transcribe`, and dictation get full-file transcripts on PyTorch backends (Windows CUDA, Linux, etc.)
- Includes unit tests for chunking and text joining

## Motivation

Whisper only accepts ~30 seconds of audio per forward pass. Without chunking, longer captures (videos, podcasts, meeting recordings) were truncated or incomplete.

## Test plan

- [x] `pytest backend/tests/test_whisper_stt_chunks.py` (3 tests pass)
- [ ] Manual: upload a >60s capture in Captures tab and verify full transcript
- [ ] Manual: `POST /transcribe` with a long audio file

## Notes

- MLX (Apple Silicon) path unchanged — MLX Whisper already accepts file paths directly
- Chunk overlap may produce minor duplicate words at window boundaries; acceptable tradeoff for full coverage

## Summary by CodeRabbit

* **New Features**
  * Added longer-audio speech recognition support with overlapping chunks for more reliable transcription.
  * Improved PyTorch-based transcription handling for better end-to-end voice input support.

* **Documentation**
  * Added a developer setup guide with local installation, startup steps, troubleshooting tips, and GPU/VRAM guidance.

* **Tests**
  * Added coverage for audio chunking and transcript merging behavior.

### DIFF
--- a/VOICEBOX_DEV_GUIDE.md
+++ b/VOICEBOX_DEV_GUIDE.md
@@ -1,0 +1,134 @@
+# Voicebox Developer Guide
+
+Local development environment for [Voicebox](https://github.com/jamiepine/voicebox) on this machine.
+
+## Environment Summary
+
+| Item | Value |
+|------|--------|
+| **Machine** | ASUS TUF A15 (Windows) |
+| **RAM** | 16 GB |
+| **GPU** | NVIDIA GeForce RTX 3060 Laptop GPU (6 GB VRAM) |
+| **Repo path** | `F:\VoiceBox\voicebox-src` |
+| **Python venv** | `backend\venv` |
+| **Backend URL** | `http://127.0.0.1:17493` |
+| **PyTorch** | CUDA 12.8 (`cu128`) — Ampere-compatible |
+| **Package manager** | Bun (JS), pip (Python), Cargo (Rust/Tauri) |
+
+### What `just setup` installs
+
+- Python virtual environment and FastAPI backend dependencies
+- CUDA-enabled PyTorch when an NVIDIA GPU is detected
+- TTS engines (Qwen3-TTS, Kokoro, Chatterbox, LuxTTS, TADA, etc.)
+- Bun workspaces for `app/`, `tauri/`, and `web/`
+- Tauri dev sidecar placeholders (`bun run setup:dev`)
+
+### Useful commands
+
+```powershell
+just --list          # all recipes
+just dev-backend     # API only
+just dev-frontend    # Tauri only (backend must be running)
+just kill            # stop dev processes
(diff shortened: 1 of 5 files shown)


priusdigital:
This addresses #626. Confirming the bug is not limited to imported
files — the live dictation path is affected too (Voicebox 0.5.0,
Windows 11, CUDA/PyTorch backend): captures over ~30s
record and play back in full, but only the first ~30s are transcribed.

Since this PR wires the chunking through PyTorchSTTBackend, it should
cover that path as well. Happy to test a build.

MasterFam:
The 30s-per-forward-pass limit is real and worth fixing, but as written the overlap is never removed, so every overlapped region is emitted twice.

`iter_whisper_chunks` steps by `CHUNK - STRIDE` (25s), so consecutive windows share 5s, and `join_whisper_chunk_texts` concatenates with a plain space. Running the helper as-is over a 90s input:

```
  0.0s ->  30.0s
 25.0s ->  55.0s  <-- overlaps the previous by 5s
 50.0s ->  80.0s  <-- overlaps the previous by 5s
 75.0s ->  90.0s  <-- overlaps the previous by 5s

4 chunks covering 105s of audio for a 90s file - 15s transcribed twice
```

Any speech inside those 5s bands is heard by two windows and appears twice in the joined text:

```
'the note said nothing at all no name no name no number just a date
 just a date by the time anyone checked anyone checked the file was gone'
```

The overlap itself is the right instinct — it stops a word being cut in half at a boundary — but it has to be used as *context* and then trimmed, rather than concatenated. That's what `transformers`' own `pipeline("automatic-speech-recognition", chunk_length_s=30, stride_length_s=5)` does: it drops the stride regions from each chunk's output using token timestamps, so boundaries are covered without duplication. Delegating to it (or reimplementing the trim with `return_timestamps=True`) would fix this and remove most of `transcribe_whisper_pytorch`.

Worth noting that the current tests can't catch it. `test_iter_whisper_chunks_long_audio_multiple_windows` asserts `sum(len(c) for c in chunks) > len(audio)`, which asserts the double-counting is present rather than that it's handled, and the join test only checks empty strings are skipped. A test that transcribes known speech spanning a boundary and asserts each word appears once would pin the behaviour.

Found this while building against the backend from outside — happy to be wrong if the duplication is trimmed somewhere downstream that I've missed.

PR #848, opened 2026-07-03
fix(macos): fullscreen dictation, correct text injection, and MLX backend crash

hussaintrawadi:
## Summary

Three independent bug fixes for macOS push-to-talk dictation, found while using
VoiceBox daily on an M4 / macOS 26. Each is a separate commit.

### 1. Dictation now works in native fullscreen apps — `fix(macos)`
The dictation pill was an ordinary `NSWindow`, which macOS never places on
another app's native fullscreen Space. In fullscreen the pill therefore never
appeared, its `WKWebView` stayed occluded, and `getUserMedia()` never resolved —
so recording only began after leaving fullscreen. The pill is now a **key‑capable
`NSPanel`** (the window class fullscreen Spaces admit; key‑capable so WebKit still
grants mic capture) with a `CanJoinAllSpaces | FullScreenAuxiliary` collection
behavior, ordered front via `orderFrontRegardless`.

### 2. Correct text injection & keyboard focus — `fix(macos)`
`capture_focus()` read the *system‑wide* focused element, which resolved to the
pill itself while it briefly held key focus. Two consequences:
- the transcript was dropped (the paste‑into‑self guard fired), so text was
  recorded but never injected — reproducible in terminals and some Electron apps;
- the pill kept keyboard focus, so the next Enter went to the pill instead of the
  user's field.

Focus now targets the **frontmost application** (always the real dictation target,
since the pill is a non‑activating panel), falls back to it for apps that expose
no accessibility focus, and the pill is restored after paste **without** re‑taking
key focus so the target keeps keyboard focus.

### 3. MLX backend crash & refinement failure — `fix(mlx)` + `fix(backend)`
- **Thread‑affinity crash:** `mlx_lm` binds a module‑level GPU stream
  (`mlx_lm/generate.py`) to whichever thread first imports it, but the server
  dispatches inference through an `asyncio.to_thread` pool. Successive calls on
  different workers hit `There is no Stream(gpu, N) in current thread` (LLM
  refinement) or, under worse timing, a hard `SIGABRT` inside
  `mlx::core::metal::get_command_encoder` that takes down the backend. All MLX
  Whisper (STT) and Qwen (LLM) load/inference now runs on a single dedicated
  worker thread, preserving stream affinity and serializing Metal.
- **Broken pipe:** the desktop host stops draining the sidecar's stdout/stderr
  after startup, closing the pipe; later writes raised `BrokenPipeError` and
  surfaced as a 500 (`[Errno 32] Broken pipe`) during transcription. The frozen
  entry point's streams now swallow pipe errors.

## Testing
- Fullscreen dictation verified end‑to‑end in Chrome and WhatsApp: pill appears
  over the fullscreen app, records, transcribes, and pastes.
- Text injection verified into desktop apps including terminals; Enter after paste
  no longer requires re‑focusing the field.
- MLX fix verified by hammering the running backend with concurrent
  prewarm/transcribe/refine — previously failed on refinement and eventually
  crashed (macOS crash report: `SIGABRT` in `get_command_encoder`); now stable
  across many rounds with no stream errors and no new crash reports.
- `cargo check` and Python syntax pass on a clean upstream checkout.

## Platform notes
Changes 1 and 2 are macOS‑only (`#[cfg(target_os = "macos")]`). Change 3 helps any
Apple‑Silicon user (the MLX crash) and every platform (the broken pipe).

## Summary by CodeRabbit

* **New Features**
  * Improved macOS dictation window behavior so it appears more reliably and works better with fullscreen apps.
  * Enhanced focus detection to better choose the correct target app when starting dictation or pasting text.

* **Bug Fixes**
  * Reduced paste failures on macOS by handling app activation more carefully.
  * Prevented logging crashes when output pipes are no longer available.
  * Improved stability for AI model loading and transcription/generation by keeping related work on a consistent execution thread.

### DIFF
--- a/backend/backends/mlx_backend.py
+++ b/backend/backends/mlx_backend.py
@@ -292,8 +292,12 @@ class MLXSTTBackend:
         if self.model is not None and self.model_size == model_size:
             return
 
-        # Run blocking load in thread pool
-        await asyncio.to_thread(self._load_model_sync, model_size)
+        # Load on the dedicated MLX thread (see services/mlx_thread.py) so the
+        # model's GPU stream stays affine to the same thread that later runs
+        # transcription.
+        from ..services.mlx_thread import run_mlx
+
+        await run_mlx(self._load_model_sync, model_size)
 
     # Alias for compatibility
     load_model = load_model_async
@@ -363,5 +367,9 @@ class MLXSTTBackend:
             else:
                 return str(result).strip()
 
-        # Run blocking transcription in thread pool
-        return await asyncio.to_thread(_transcribe_sync)
+        # Run transcription on the dedicated MLX thread (see
+        # services/mlx_thread.py) — same thread as the model load, so MLX's
+        # thread-local GPU streams resolve.
+        from ..services.mlx_thread import run_mlx
+
+        return await run_mlx(_transcribe_sync)
--- a/backend/backends/qwen_llm_backend.py
+++ b/backend/backends/qwen_llm_backend.py
@@ -212,7 +212,9 @@ class MLXQwenLLMBackend:
         if self.model is not None and self._current_model_size != model_size:
             self.unload_model()
 
-        await asyncio.to_thread(self._load_model_sync, model_size)
+        from ..services.mlx_thread import run_mlx
+
+        await run_mlx(self._load_model_sync, model_size)
 
     def _load_model_sync(self, model_size: str) -> None:
         from mlx_lm import load as mlx_load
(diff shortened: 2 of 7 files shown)


leighton-tidwell:
If we could get this merged, it would be huge. This is going to tackle three bugs that are effecting me every day and preventing me from using this as my main driver.

leighton-tidwell:
This PR fixes **two** separate problems we hit daily on **Voicebox 0.5.0** (macOS 26.5.1, Apple Silicon):

1. **Text injection / auto-paste into terminals & Electron apps** — capture + transcription work, but the transcript is never pasted into e.g. the Codex CLI / terminal. Your "paste-into-self guard fired, text recorded but never injected — reproducible in terminals and some Electron apps" is exactly our symptom.
2. **MLX refinement crash** — `There is no Stream(gpu, 1) in current thread.` (repro'd with `mlx-community/Qwen3-4B-4bit`).

High-value PR — it resolves both for us. Happy to test a build.

PR #862, opened 2026-07-08
fix(docker): exempt rocm-entrypoint.sh from .dockerignore

### LINKED ISSUE #835: docker compose fails due to scripts/rocm-entrypoint.sh being excluded via the .dockerignore file
Docker Compose did not work at first for me, because it tries to copy the script "rocm-entrypoint.sh", but can't find it because the "scripts/" directory is listed in the .dockerignore file. I fixed this problem by simply adding this line to the .dockerignore file:

!scripts/rocm-entrypoint.sh

Xarianne:
## Problem

The `scripts/` directory is blanket-excluded in `.dockerignore`, which prevents `scripts/rocm-entrypoint.sh` from reaching the Docker build context. The Dockerfile's final stage copies this file as the container entrypoint:

```dockerfile
COPY --chmod=755 scripts/rocm-entrypoint.sh /usr/local/bin/entrypoint.sh
```

So the ROCm build fails with:

```
failed to solve: failed to compute cache key: failed to calculate checksum of ref
...: "/scripts/rocm-entrypoint.sh": not found
```

## Fix

Add a negation rule in `.dockerignore` to let the entrypoint script through while keeping the rest of `scripts/` excluded:

```diff
 scripts/
+!scripts/rocm-entrypoint.sh
```

Fixes #835

## Summary by CodeRabbit

* **Chores**
  * Updated Docker-related packaging rules to ensure the required startup script remains included in container builds.
* **Bug Fixes**
  * Improved GPU/ROCm startup behavior by clearing an empty `HSA_OVERRIDE_GFX_VERSION` value so it no longer interferes with GPU detection.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -9,6 +9,7 @@ landing/
 docs/
 mlx-test/
 scripts/
+!scripts/rocm-entrypoint.sh
 
 # Dependencies & build artifacts (rebuilt in Docker)
 node_modules/
--- a/backend/app.py
+++ b/backend/app.py
@@ -38,6 +38,13 @@ logging.basicConfig(
 
 logger = logging.getLogger(__name__)
 
+# An empty HSA_OVERRIDE_GFX_VERSION poisons the ROCm HSA runtime — it's
+# treated as "force-empty" and no GPU is detected, even natively supported
+# ones (e.g. gfx1201 / RX 9070 on ROCm 7.2). docker-compose can't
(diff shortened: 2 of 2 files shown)


Xarianne:
It looks like another PR was submitted almost at the same time for dockerignore: #861 

So part of my PR does the same thing. I am going to close this issue and open a new one just with the app.py fix.

Issue #680, opened 2026-05-20
[Linux] run `just dev` crashed on startup with `Option::unwrap()` panic in tao event loop


fweng322:
Since there's no Linux package now, I checkout the git repository and built on my (k)ubuntu 24.04.

`just build` succeeded. But when running `just dev` (or `cd tauri && bun run tauri dev`) brings up the main window and then immediately aborts the process with a non-unwinding panic from `tao` before the UI is usable. 

## Environment

- OS: Linux 6.8.0-117-generic (Ubuntu-flavoured), x86_64
- Voicebox: 0.5.0 (branch `main`, post #544)
- Rust toolchain: stable (cargo registry shows `tao 0.34.5`, `tauri 2.9.5`, `tauri-runtime-wry 2.9.3`, `glib 0.18.5`, `gtk 0.18.2`)
- Display server: X11 / GTK 3 via WebKitGTK
- Python: 3.12.3 (uvicorn backend on :17493)

## Reproduction

1. `just setup` (one-time).
2. `just dev`.
3. Wait for cargo to finish building `voicebox` and the binary to launch (`Running target/debug/voicebox`).
4. The main window briefly opens, then the process aborts.

The frontend at this point has done nothing more than load the SPA — the only HTTP request that reaches the backend before the crash is `GET /events/speak` (the Rust-side SSE subscriber in `speak_monitor`).

## Expected

`just dev` brings up the desktop app and stays running, mirroring `just build`'s bundled behaviour.

## Actual

It crashed on startup.  Console error log:

```
INFO:     127.0.0.1:36638 - "GET /events/speak HTTP/1.1" 200 OK


thread 'main' (18104) panicked at /home/franklin/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tao-0.34.5/src/
platform_impl/linux/event_loop.rs:449:18:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

thread 'main' (18104) panicked at /rustc/59807616e1fa2540724bfbac14d7976d7e4a3860/library/core/src/panicking.rs:225:5
:
panic in a function that cannot unwind
stack backtrace:
INFO:     127.0.0.1:33184 - "OPTIONS /settings/captures HTTP/1.1" 200 OK
   0:     0x5cbba2d390ba - std[e28293b1aa0f68bd]::backtrace_rs::backtrace::libunwind::trace
[...]

PR #875, opened 2026-07-10
Fixes #873 by avoiding the `tao` Linux/X11 panic triggered when `set_ignore_cursor_events()` is called on the transparent dictate overlay window.

### LINKED ISSUE #873: tao panic on Linux  set_ignore_cursor_events crashes on transparent windows
Summary
After successful compilation, the Tauri app launches, renders the UI, serves API requests, and then immediately crashes with a panic in the tao windowing library. The crash is 100% reproducible on every launch.

Error

thread 'main' panicked at tao-0.34.5/src/platform_impl/linux/event_loop.rs:449:18:
called `Option::unwrap()` on a `None` value
Root cause
The dictate overlay window is created with .transparent(true). On Linux/X11, transparent windows don't always create a proper GDK window internally. When the app calls window.set_ignore_cursor_events(true) (line 1424 in main.rs), tao internally does:

rust

// tao-0.34.5/src/platform_impl/linux/event_loop.rs:448-449
window
  .window()    // Returns None for transparent windows on X11
  .unwrap()    // PANIC!
  .input_shape_combine_region(&empty_region, 0, 0);
The window.window() call returns None because the GDK window hasn't been realized for the transparent overlay, and the .unwrap() panics.

Environment
OS: Ubuntu 24.04 (X11, DISPLAY=:1)
GPU: NVIDIA GeForce RTX 3050 A Laptop GPU
Rust: 1.96.0
tao: 0.34.5
tauri: 2.9.5

Upstream
This is arguably a bug in tao itself — window.window() should be guarded instead of blindly .unwrap()'d. An upstream issue at [tauri-apps/tao](https://github.com/tauri-apps/tao) may also be appropriate.

krishna3554:
## Summary

Fixes #873 by avoiding the `tao` Linux/X11 panic triggered when `set_ignore_cursor_events()` is called on the transparent dictate overlay window.

## Changes

- Disable dictate window transparency on Linux.
- Skip `set_ignore_cursor_events(true)` in the `dictate:hide` handler on Linux.
- Skip `set_ignore_cursor_events(false)` in dictate show paths on Linux.
- Preserve existing behavior on macOS and Windows.

## Why

On Linux/X11, transparent Tauri windows can leave tao without a realized GDK window. Calling `set_ignore_cursor_events()` then panics inside tao when it unwraps a missing window handle.

On Linux, the dictate overlay is already moved off-screen and hidden when dismissed, so skipping click-through toggling avoids the crash without leaving an interactive invisible window.

## Verification

- `cargo check` was attempted but could not complete because the local checkout is missing the expected Tauri sidecar binary:
  `binaries/voicebox-server-x86_64-unknown-linux-gnu`
- `cargo fmt --check` was attempted, but the repository currently has pre-existing formatting differences across unrelated Rust files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Linux stability when using the dictate window by preventing crashes related to click-through behavior.
  * Preserved cursor interaction behavior on supported platforms while avoiding unsupported Linux window handling.

### DIFF
--- a/tauri/src-tauri/src/hotkey_monitor.rs
+++ b/tauri/src-tauri/src/hotkey_monitor.rs
@@ -264,7 +264,10 @@ fn apply_effect(app: &AppHandle, effect: Effect) {
                         let _ = window.set_position(tauri::PhysicalPosition::new(x, y));
                     }
                 }
-                let _ = window.set_ignore_cursor_events(false);
+                #[cfg(not(target_os = "linux"))]
+                {
+                    let _ = window.set_ignore_cursor_events(false);
+                }
                 // Deliberately no set_focus() — taking key focus would yank
                 // it out of whatever app the user was typing in, which is
                 // the opposite of what a dictation overlay should do.
--- a/tauri/src-tauri/src/main.rs
+++ b/tauri/src-tauri/src/main.rs
@@ -38,7 +38,11 @@ fn build_dictate_window(app: &tauri::AppHandle) -> tauri::Result<tauri::WebviewW
     .title("Voicebox Dictate")
     .inner_size(DICTATE_WINDOW_WIDTH, DICTATE_WINDOW_HEIGHT)
     .decorations(false)
-    .transparent(true)
+    // Transparent windows can cause tao to panic on Linux/X11 when
+    // set_ignore_cursor_events is called because the underlying GDK window
+    // may not be realized. Keep transparency for platforms where the
+    // click-through workaround is supported.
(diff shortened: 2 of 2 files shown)


jamiepine:
Thanks for taking a run at the dictate-overlay panic — it's one of the most-reported Linux issues (#680/#873). Of the three PRs addressing it we're going with #927's approach, which handles the unrealized-window case at the source rather than at the call site, so closing this one. Appreciate the repro work.

Issue #723, opened 2026-05-31
MLX backend uses CPU variant on Apple Silicon despite gpu_available=true (M5 Pro)


Petrusreno:
## Environment
- macOS Darwin 25.5.0, Apple M5 Pro, 24 GB
- Voicebox v0.5.0 (DMG `Voicebox_0.5.0_aarch64.dmg`)
- Model: Qwen3-TTS-12Hz-1.7B-Base (loaded)
- Settings → GPU enabled

## Symptom
`GET /health` correctly detects MPS but reports `backend_variant: cpu`:

```json
{
  "status": "healthy",
  "model_loaded": true,
  "model_size": "1.7B",
  "gpu_available": true,
  "gpu_type": "MPS (Apple Silicon)",
  "backend_type": "mlx",
  "backend_variant": "cpu",
  "gpu_compatibility_warning": null
}
```

`~/Library/Application Support/sh.voicebox.app/backends/` contains only an empty `cuda/` subfolder — no MPS-variant directory shipped.

## Performance impact
End-to-end test through `POST /generate` (Petrus PT-BR cloned profile, ~80 chars input):
- Wall-clock to `status=completed`: **20.1 s**
- Audio duration: **5.44 s**
- **RTF: 3.7** (vs. Qwen3-TTS upstream claim of sub-realtime via Metal/MLX)

`POST /generate/stream` is non-streaming in practice — it returns audio/wav only after the full clip is rendered (TTFB = total time, 15.5 s in another test).

## Repro
```bash
curl -X POST http://127.0.0.1:17493/generate \
  -H "Content-Type: application/json" \
  -d '{"profile_id":"<your-petrus-id>","text":"Bom dia, doutor. Tudo bem?","language":"pt","engine":"qwen","model_size":"1.7B"}'
# poll GET /history/{id} until status=completed
```

## Expected
On Apple Silicon, MLX should use the Metal backend (typically 5-10× faster than CPU for autoregressive LMs of this size). Qwen3-TTS reports ~97 ms TTFB and sub-realtime RTF on properly configured Apple Silicon stacks.

## Suggested
- Ship MLX-Metal variant alongside MLX-CPU in the `backends/` distribution
- OR: expose a runtime override so users can pull the Metal variant on first launch
- Bonus: make `/generate/stream` actually stream chunks as they're produced (Qwen3-TTS supports incremental output)

[...]

Issue #706, opened 2026-05-25
MLX falls back to CPU on Apple M5 Max — GPU not utilized (voicebox 0.5.0)


dycx:
## Environment

- Voicebox version: 0.5.0
- macOS: 26.5
- Hardware: MacBook Pro, Apple M5 Max
- Network: direct connection (no proxy)

## Symptoms

The `/health` endpoint reports:

```json
{
  "backend_type": "mlx",
  "backend_variant": "cpu",
  "gpu_available": true,
  "gpu_type": "MPS (Apple Silicon)"
}
```

MLX is selected as the backend, but falls through to CPU variant despite M5 Max having available Metal GPU. Generation succeeds but is extremely slow (expected CPU-only speeds rather than the 4-5× MLX Metal acceleration).

## Expected behavior

MLX backend should use Metal GPU (`backend_variant: "gpu"`) on M5 Max, same as M1–M4 series.

## Possible cause

Similar to ollama/ollama#15642 — the M5 Max Metal GPU family may not be recognized by the bundled MLX version, preventing selection between `mlx_metal_v3` / `mlx_metal_v4` and defaulting to CPU.

## Steps to reproduce

1. Install Voicebox 0.5.0 on Apple M5 Max running macOS 26
2. Download Qwen3-TTS 1.7B model
3. Generate speech via UI or API
4. Check `/health` — shows `backend_variant: "cpu"

harukaxxxx:
My M4 Mac mini is also experiencing slow generation. I checked the /health endpoint and confirmed that my GPU is not being utilized either. It keeps falling back to the CPU.

Here is my /health output:

```
{
  "status": "healthy",
  "model_loaded": true,
  "model_downloaded": true,
  "model_size": "0.6B",
  "gpu_available": true,
  "gpu_type": "MPS (Apple Silicon)",
  "vram_used_mb": null,
  "backend_type": "mlx",
  "backend_variant": "cpu",
  "gpu_compatibility_warning": null
}
```

Even though "gpu_available" is true, the "backend_variant" shows "cpu". It seems this MLX fallback issue isn't exclusive to the M5 Max, but also affects the M4 series.

chubarevgeniy:
Same for m3 max

jeremyforest:
Same on M3 laptop

PR #886, opened 2026-07-12
fix: MLX thread affinity crash + missing mlx-lm/mlx-audio in dev setup

### LINKED ISSUE #699: Qwen TTS generation fails 100% with `Stream(gpu, 1) not in current thread` on Apple Silicon (v0.5.0)
## Environment

- Voicebox **v0.5.0** (macOS .app bundle, PyInstaller-packaged)
- macOS Darwin 25.4.0, arm64 (Apple Silicon, 128 GB RAM)
- Python 3.12.10 (bundled)
- `mlx==0.31.2`, `mlx_audio==0.4.1`, `qwen_tts==0.1.1`
- Model: `mlx-community/Qwen3-TTS-12Hz-1.7B-Base-bf16` (downloaded, 4.2 GB)
- `/health` reports: `backend_type: mlx`, `backend_variant: cpu`, `gpu_type: MPS (Apple Silicon)`, `model_loaded: true`, `gpu_available: true`

## Symptom

Every generation fails **immediately** with:

```
There is no Stream(gpu, 1) in current thread.
```

The UI stays on "Loading model..." but the SQLite row in `voicebox.db` is already `status=failed`. No Python traceback is written to the server log — only the message is captured in the `generations.error` column.

## Reproduction (100% repeatable)

After a clean app start (`Ready` logged, model loaded), call the API directly:

```bash
curl -X POST http://127.0.0.1:17493/generate \
  -H 'Content-Type: application/json' \
  -d '{"profile_id":"<any-existing-profile>","text":"hello","language":"en","engine":"qwen","model_size":"1.7B"}'
```

Polling `/generate/{id}/status` flips to `failed` within ~1 second, every time. Confirmed across multiple profiles, languages (zh/en), and short/long inputs.

## Likely root cause

`mlx_audio/stt/generate.py:224` (and almost certainly similar code on the TTS path):

```python
generation_stream = mx.new_stream(mx.default_device())
```

The stream is created on the main/import thread, but inference runs in a uvicorn worker thread (`run_in_executor` / `asyncio.to_thread`). MLX streams are **thread-local**, so the worker thread sees no `Stream(gpu, 1)` and raises.

Fix direction: either recreate the stream inside the worker thread, or use [`mlx.core.new_thread_local_stream`](https://ml-explore.github.io/mlx/build/html/python/_autosummary/mlx.core.new_thread_local_stream.html) (added in MLX 0.31).

## Suggested follow-ups

1. Move `mx.new_stream` into the worker that actually runs inference, or switch to `new_thread_local_stream`.
2. Let the exception traceback through to the server log on generation failure — currently the message is silently captured in SQLite only, which makes the issue look like "model still loading" in the UI.
3. The `backend_variant: cpu` while `backend_type: mlx` and `gpu_type: MPS` reported by `/health` looks inconsistent and is worth a sanity check.

## Workaround

None found from the user side — backend cannot be switched away from MLX in the bundled build, and the error is raised before any inference happens.

### LINKED ISSUE #823: just setup-python doesn't install mlx-audio/mlx-lm — MLX backend fails on fresh Apple Silicon dev setup
## Summary

`just setup-python` does not install `mlx-audio` (or `mlx-lm`), so a freshly set-up dev environment on Apple Silicon fails **all** STT/TTS that go through the MLX backend with:

```
ModuleNotFoundError: No module named 'mlx_audio'
```

The release CI (`.github/workflows/release.yml`) installs these packages, but the `justfile` — the documented dev setup path (`just setup`) — does not. So the bundled app works while a from-source dev checkout silently can't transcribe or generate on the MLX path.

## Environment

- Repo at `cabef1b` (v0.5.0), built from source via `just setup`
- macOS 26.5.1, Darwin 25.5.0, arm64 (Apple Silicon)
- Python 3.12.13 (Homebrew), `mlx==0.31.2`

## Reproduction

```bash
just setup           # or: just setup-python
just dev-backend
curl -X POST http://127.0.0.1:17493/transcribe -F "file=@any.wav" -F "model=turbo"
```

The transcribe call returns `202` repeatedly, and the server log shows:

```
ERROR: Marked whisper-turbo as error: No module named 'mlx_audio'
```

`backend/backends/mlx_backend.py` imports `from mlx_audio.stt import load` (line ~307) and `from mlx_audio.tts import load` (line ~97), but nothing in `just setup-python` installs `mlx-audio`.

## Root cause

`requirements-mlx.txt` deliberately does **not** list `mlx-audio` (its comment explains the `transformers` version conflict and says to install it `--no-deps` afterward). `release.yml` does exactly that:

```yaml
# .github/workflows/release.yml
pip install -r backend/requirements-mlx.txt
pip install --no-deps mlx-lm==0.31.1
pip install --no-deps mlx-audio==0.4.1
```

But the `justfile` MLX block stops at `requirements-mlx.txt`:

```make
# justfile (Apple Silicon block in setup-python)
if [ "$(uname -m)" = "arm64" ] && [ "$(uname)" = "Darwin" ]; then
    echo "Detected Apple Silicon — installing MLX dependencies..."
    {{ pip }} install -r {{ backend_dir }}/requirements-mlx.txt
fi
```

So `mlx-lm` and `mlx-audio` are never installed for dev.

## Suggested fix

Add the same two `--no-deps` installs to the Apple-Silicon branch of `setup-python` in the `justfile` (both the `[unix]` and `[windows]` recipes), mirroring `release.yml`:

```make
{{ pip }} install -r {{ backend_dir }}/requirements-mlx.txt
{{ pip }} install --no-deps mlx-lm==0.31.1
{{ pip }} install --no-deps mlx-audio==0.4.1
```

## Workaround

```bash
backend/venv/bin/pip install --no-deps mlx-lm==0.31.1 mlx-audio==0.4.1
```

---

*Filed after a from-source build on Apple Silicon. Related: #699 (also MLX-on-Apple-Silicon, same package versions).*

AlexStephenLytton:
## Summary

Two related bugs on Apple Silicon that turned out to share a root cause, plus one cosmetic health-endpoint fix found while chasing them down.

- Fixes #699 (Qwen TTS generation fails 100% with `Stream(gpu, 1) not in current thread`)
- Fixes #823 (`just setup-python` doesn't install mlx-audio/mlx-lm — MLX backend fails on fresh Apple Silicon dev setup)
- Related to #706 / #723 (MLX falls back to CPU / `backend_variant` misreports) — the health.py fix here addresses the *reporting* side of those; see notes below on what it does and doesn't cover.

## What's actually going on

Starting from a clean `just setup-python` on Apple Silicon, MLX never actually gets used:

1. `requirements-mlx.txt` only lists `mlx` + `miniaudio`. `just setup-python` installs it and stops — it never runs the `pip install --no-deps mlx-lm==0.31.1` / `mlx-audio==0.4.1` steps that `.github/workflows/release.yml` already does for packaged builds. Without `mlx_audio` importable, `get_backend_type()` silently falls back to `"pytorch"` — no error, no log line calling it out, health endpoint still looks plausible. This is #823.

2. The obvious fix once you notice #823 is `pip install mlx-lm` by hand. That reintroduces a *second* problem: `mlx-lm>=0.30.0` requires `transformers>=5.0.0`, conflicting with the `transformers<=4.57.6` cap in `requirements.txt`. `transformers>=5.0` changed `check_model_inputs` from a decorator factory (`@check_model_inputs()`) to a plain decorator (`@check_model_inputs`, i.e. `check_model_inputs(func)`). `qwen_tts`'s CustomVoice tokenizer (`qwen_tts/core/tokenizer_12hz/modeling_qwen3_tts_tokenizer_v2.py`) still calls the old convention, so once transformers gets silently bumped, loading Qwen CustomVoice raises:
   ```
   TypeError: check_model_inputs() missing 1 required positional argument: 'func'
   ```
   `--no-deps` avoids this entirely (mlx-lm's own transformers requirement is never consulted, so transformers stays at whatever `requirements.txt` installed) — which is exactly what `release.yml` already does, it just never made it into `justfile`.

3. Once mlx-audio *is* installed and the backend actually reaches MLX, every model load/generate/transcribe call runs through `asyncio.to_thread()`, which uses the event loop's default thread pool — no guarantee that two calls land on the same OS thread. MLX's Metal command encoder/stream is thread-local (bound to whichever thread first touched the GPU device), so if model load and a later `generate()` land on different pool threads, MLX raises:
   ```
   There is no Stream(gpu, 1) in current thread.
   ```
   This is #699, reproducible 100% of the time per the original report. Fix: pin every MLX call to one dedicated worker thread instead of the shared pool.

4. Along the way: `/health`'s `backend_variant` field never had a branch for `backend_type == "mlx"`, so it always reported `"cpu"` regardless of whether MLX was actually active — this is what #706 and #723 point at as evidence MLX wasn't using the GPU. Added the missing branch (`"mps"`). This is a reporting fix only; it doesn't address whatever's causing MLX-Metal generation to be slower than PyTorch-CPU on the hardware in those two reports specifically — I don't have that hardware to reproduce on, so I'm not claiming this closes them, just that the `backend_variant` field itself was lying.

## Changes

- `justfile`: add the two missing `pip install --no-deps` lines to the Unix `setup-python` recipe, mirroring `release.yml` exactly.
- `backend/requirements-mlx.txt`: add `sounddevice==0.5.3` (mlx-audio's own pin, previously unresolved on a from-scratch install), and a NOTE explaining why `mlx-lm`/`mlx-audio` are deliberately *not* listed here and must go through the `--no-deps` path instead.
- `backend/backends/mlx_backend.py`: dedicated single-worker `ThreadPoolExecutor` for all MLX calls (`_run_on_mlx_thread`), replacing the 4 `asyncio.to_thread()` call sites across `MLXTTSBackend` and `MLXSTTBackend`.
- `backend/routes/health.py`: 2-line addition — `elif backend_type == "mlx": default_variant = "mps"`.

## Testing

All verified on real Apple Silicon hardware (M4):

- Clean venv, ran `just setup-python` equivalent commands by hand (matching the justfile diff) → `mlx_audio` importable, `/health` reports `backend_type: "mlx"`, `backend_variant: "mps"` (previously `"pytorch"` / `"cpu"`).
- Multiple back-to-back `qwen` engine generations (model load + several `generate()` calls across different requests) with no `Stream(gpu, N)` crash.
- Qwen CustomVoice 0.6B: `pip install mlx-lm==0.31.1` (no `--no-deps`, reproducing the trap) → confirmed the exact `check_model_inputs()` TypeError from the report above. Reinstalled per this PR's `--no-deps` sequence → CustomVoice downloads, loads, and generates successfully.
- Regression-tested the base `qwen` MLX engine after the transformers fix to make sure it still works (it does — mlx_audio.tts doesn't touch qwen_tts at all, only the CustomVoice/PyTorch code path does).

Happy to adjust anything — this is my first PR against this repo, let me know if there's a preferred format for splitting these into separate PRs instead of one.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved MLX speech-to-text and text-to-speech stability on Apple Silicon by running blocking MLX tasks through a dedicated worker.
  - Prevented unintended network activity during MLX model/cache operations.
  - Fixed health check variant selection to correctly surface the MPS variant for the MLX backend.
  - Expanded device selection to use Apple MPS for PyTorch and Qwen custom voice backends, including clearing MPS cache during unload.

- **Chores**
  - Updated Apple Silicon dependency setup by pinning MLX backend packages (including audio) and the required sounddevice runtime.

### DIFF
--- a/backend/backends/base.py
+++ b/backend/backends/base.py
@@ -171,10 +171,10 @@ def check_cuda_compatibility() -> tuple[bool, str | None]:
 
 def empty_device_cache(device: str) -> None:
     """
-    Free cached memory on the given device (CUDA or XPU).
+    Free cached memory on the given device (CUDA, XPU, or MPS).
 
-    Backends should call this after unloading models so VRAM is returned
-    to the OS.
+    Backends should call this after unloading models so VRAM/unified
+    memory is returned to the OS.
     """
     import torch
 
@@ -182,6 +182,8 @@ def empty_device_cache(device: str) -> None:
         torch.cuda.empty_cache()
     elif device == "xpu" and hasattr(torch, "xpu"):
         torch.xpu.empty_cache()
+    elif device == "mps" and hasattr(torch, "mps"):
+        torch.mps.empty_cache()
 
 
 def manual_seed(seed: int, device: str) -> None:
--- a/backend/backends/mlx_backend.py
+++ b/backend/backends/mlx_backend.py
@@ -5,10 +5,26 @@ MLX backend implementation for TTS and STT using mlx-audio.
 from typing import Optional, List, Tuple
 import asyncio
 import logging
+from concurrent.futures import ThreadPoolExecutor
 import numpy as np
 from pathlib import Path
 
 logger = logging.getLogger(__name__)
+
+# MLX's Metal command encoder/stream is thread-local: it's bound to
+# whichever thread first touches the GPU device. asyncio.to_thread() uses
+# the loop's default executor, which can (and does) hand different calls
+# to different worker threads — model load on one thread, then generate()
+# on another raises "There is no Stream(gpu, N) in current thread." Pin
+# every MLX call to the same single dedicated worker instead.
+# See https://github.com/jamiepine/voicebox/issues/699
+_mlx_executor = ThreadPoolExecutor(max_workers=1, thread_name_prefix="mlx-worker")
+
+
+def _run_on_mlx_thread(func, *args):
+    """Run ``func(*args)`` on the single dedicated MLX worker thread."""
+    loop = asyncio.get_running_loop()
+    return loop.run_in_executor(_mlx_executor, func, *args)
 
 # PATCH: Import and apply offline patch BEFORE any huggingface_hub usage
 # This prevents mlx_audio from making network requests when models are cached
@@ -81,8 +97,8 @@ class MLXTTSBackend:
         if self.model is not None and self._current_model_size != model_size:
             self.unload_model()
 
-        # Run blocking load in thread pool
-        await asyncio.to_thread(self._load_model_sync, model_size)
+        # Run blocking load on the dedicated MLX thread
+        await _run_on_mlx_thread(self._load_model_sync, model_size)
 
     # Alias for compatibility
     load_model = load_model_async
@@ -258,8 +274,8 @@ class MLXTTSBackend:
 
             return audio, sample_rate
 
-        # Run blocking inference in thread pool
-        audio, sample_rate = await asyncio.to_thread(_generate_sync)
+        # Run blocking inference on the dedicated MLX thread
+        audio, sample_rate = await _run_on_mlx_thread(_generate_sync)
 
         return audio, sample_rate
 
@@ -292,8 +308,8 @@ class MLXSTTBackend:
         if self.model is not None and self.model_size == model_size:
             return
 
-        # Run blocking load in thread pool
-        await asyncio.to_thread(self._load_model_sync, model_size)
+        # Run blocking load on the dedicated MLX thread
+        await _run_on_mlx_thread(self._load_model_sync, model_size)
 
     # Alias for compatibility
     load_model = load_model_async
@@ -363,5 +379,5 @@ class MLXSTTBackend:
             else:
                 return str(result).strip()
 
-        # Run blocking transcription in thread pool
-        return await asyncio.to_thread(_transcribe_sync)
+        # Run blocking transcription on the dedicated MLX thread
+        return await _run_on_mlx_thread(_transcribe_sync)
--- a/backend/backends/mlx_backend.py
+++ b/backend/backends/mlx_backend.py
@@ -93,9 +93,11 @@ class MLXTTSBackend:
         if self.model is not None and self._current_model_size == model_size:
             return
 
-        # Unload existing model if different size requested
+        # Unload existing model if different size requested — routed
+        # through the same dedicated thread as load/generate so it stays
+        # serialized with any in-flight MLX call (see _run_on_mlx_thread).
         if self.model is not None and self._current_model_size != model_size:
-            self.unload_model()
+            await _run_on_mlx_thread(self.unload_model)
 
         # Run blocking load on the dedicated MLX thread
         await _run_on_mlx_thread(self._load_model_sync, model_size)
--- a/backend/backends/pytorch_backend.py
+++ b/backend/backends/pytorch_backend.py
@@ -34,7 +34,7 @@ class PyTorchTTSBackend:
 
     def _get_device(self) -> str:
         """Get the best available device."""
-        return get_torch_device(allow_xpu=True, allow_directml=True)
+        return get_torch_device(allow_xpu=True, allow_directml=True, allow_mps=True)
 
     def is_loaded(self) -> bool:
         """Check if model is loaded."""
@@ -256,7 +256,7 @@ class PyTorchSTTBackend:
 
     def _get_device(self) -> str:
         """Get the best available device."""
-        return get_torch_device(allow_xpu=True, allow_directml=True)
+        return get_torch_device(allow_xpu=True, allow_directml=True, allow_mps=True)
 
     def is_loaded(self) -> bool:
         """Check if model is loaded."""
--- a/backend/backends/qwen_custom_voice_backend.py
+++ b/backend/backends/qwen_custom_voice_backend.py
@@ -65,7 +65,7 @@ class QwenCustomVoiceBackend:
         self._current_model_size: Optional[str] = None
 
     def _get_device(self) -> str:
-        return get_torch_device(allow_xpu=True, allow_directml=True)
+        return get_torch_device(allow_xpu=True, allow_directml=True, allow_mps=True)
 
     def is_loaded(self) -> bool:
         return self.model is not None
--- a/backend/backends/qwen_custom_voice_backend.py
+++ b/backend/backends/qwen_custom_voice_backend.py
@@ -25,6 +25,7 @@ from . import TTSBackend, LANGUAGE_CODE_TO_NAME
 from .base import (
     is_model_cached,
     get_torch_device,
+    empty_device_cache,
     combine_voice_prompts as _combine_voice_prompts,
     model_load_progress,
 )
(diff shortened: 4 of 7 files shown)


AlexStephenLytton:
Added a second, related commit: enabling MPS for the PyTorch Qwen3-TTS/CustomVoice engines (\`allow_mps=True\` was already used elsewhere — \`luxtts_backend.py\`, \`qwen_llm_backend.py\` — just missing here).

While testing the MLX thread fix I benchmarked PyTorch+MPS directly against MLX for the same engine/model/machine (M4, \`qwen\` engine, 0.6B), out of curiosity about whether MLX's "4-5x faster than PyTorch" claim held up now that both paths actually work:

| Backend | RTF |
|---|---|
| MLX (this PR's thread fix) | ~3.4-4.3x |
| PyTorch + MPS (this commit) | ~1.7x |
| PyTorch + CUDA, RTX 2080 Ti, for reference | ~2.3-3.1x |

PyTorch+MPS came out faster than MLX on the same hardware, and faster than a discrete CUDA GPU too. Small sample (one text, one model size, one run each) so take the exact numbers with a grain of salt, but the gap is large enough that it seemed worth surfacing rather than just leaving PyTorch+MPS unreachable. Not proposing changing default backend selection — MLX-Whisper is reportedly faster than PyTorch-CPU for STT per the docs, so the right choice may differ per engine/task — just that MPS should be an available option instead of silently falling back to CPU whenever the PyTorch path is used.

shsunmoonlee:
Confirming this fix on more hardware, plus one gap found while testing it.

**Confirmation** — Apple M3 Pro, 18 GB, macOS 26.5 (Darwin 25.5.0), dev checkout (Python 3.12.8, `mlx 0.32.0`, `mlx-lm 0.31.1`, `mlx-audio 0.4.1`):

- Before this PR: `POST /generate` (Qwen3-TTS 1.7B) failed 100% with `There is no Stream(gpu, 0) in current thread.`
- With this PR: TTS generation and Whisper Turbo `/transcribe` both work, repeated calls stable. Also hit #823 first on a fresh `just setup` exactly as described, and the two `--no-deps` installs here resolved it — `/health` now correctly reports `backend_variant: mps` with the reporting fix.

**Gap** — the same thread-affinity crash still exists in `backend/backends/qwen_llm_backend.py`: `MLXQwenLLMBackend.load_model`/`generate` still go through `asyncio.to_thread()`, so with this PR applied, `POST /llm/generate` (personality Compose/Rewrite, dictation refinement) fails 100% with the identical error, raised from `mlx_lm.generate`'s `wired_limit` on exit.

Opened #909 (built on top of this branch) routing the LLM backend through the same `_run_on_mlx_thread` helper — verified on the hardware above. @AlexStephenLytton feel free to pick that commit into this PR instead if you'd rather keep the fix in one place.

leighton-tidwell:
Affected on **Voicebox 0.5.0** / macOS 26.5.1 / Apple Silicon. Same MLX thread-affinity crash (`There is no Stream(gpu, N) in current thread`). Note the LLM-refinement path still uses `asyncio.to_thread` in `qwen_llm_backend.py` even with this applied — #909 extends this fix to that backend, which is where our crash surfaces. Confirming the base fix here is needed; happy to test.

jamiepine:
Thanks for this — two of the three things here were real. The thread-affinity crash is fixed on `0.6-dictation-fixes` via `backend/services/mlx_thread.py` (single-worker executor, load+infer as one job), which supersedes the threading half. And the `just setup-python` gap — mlx-lm/mlx-audio not installing on Apple Silicon so fresh dev setups silently fall back to PyTorch — is a genuine catch we're tracking for 0.6.

The part we can't take is the MPS device-policy change bundled in with the fix; that alters generation behavior beyond the crash being fixed and would need its own PR with its own justification. Closing this one, but the setup-python find is appreciated.

Issue #505, opened 2026-04-20
miniaudio is missing


verhas:
On a fresh MacOS M1 install, transcribing fails because it does not find the miniaudio dependency. Probably missing from the package. The workaround is to `pip install` it globally (user-level) and restart the application. It seemed to work at first, but it does not. I can provide the text, pasting it though.

jamiepine:
Looks like it might have been accidentally removed, checking now

PR #892, opened 2026-07-13
fix(setup): install mlx-lm and mlx-audio in setup-python on Apple Silicon

flowbrix:
## Problem

On a fresh Apple Silicon dev setup, `just setup` installs `backend/requirements-mlx.txt` but not `mlx-audio`/`mlx-lm` themselves, so the Whisper MLX STT path is broken out of the box: `POST /transcribe` fails with `No module named 'mlx_audio'` (and, once that is installed, `No module named 'mlx_lm'`).

These two packages are intentionally absent from `requirements-mlx.txt` because from `mlx-audio` 0.3.1 / `mlx-lm` 0.31.1 onward they declare `transformers>=5.x`, which conflicts with the `transformers<=4.57.x` cap in `requirements.txt` (see the existing note in `backend/requirements-mlx.txt`). The release workflow (`.github/workflows/release.yml`) already handles this by installing both with `--no-deps` — but the dev setup recipe never got the same treatment.

## Fix

Mirror the release workflow in the Apple Silicon branch of the `setup-python` recipe (`[unix]` section; `[windows]` is unaffected since MLX is macOS-only):

- `pip install --no-deps mlx-lm==0.31.1`
- `pip install --no-deps mlx-audio==0.4.1`

with a comment explaining the `--no-deps` (transformers conflict), plus a pointer to the justfile in the `requirements-mlx.txt` note. Pins match `release.yml` so dev and CI stay in sync.

## Verification

Ran `just setup-python` in a clean checkout (no pre-existing venv) on an Apple Silicon Mac — completed successfully, and:

```
$ backend/venv/bin/python -c "import mlx_audio, mlx_lm"   # OK
$ backend/venv/bin/python -c "import mlx_audio.stt, mlx_audio.tts"   # OK
```

with `mlx-audio 0.4.1` / `mlx-lm 0.31.1` resolved as pinned.

## Summary by CodeRabbit

* **New Features**
  * Expanded Apple Silicon (macOS arm64) MLX setup to include required audio and language packages.
* **Documentation**
  * Updated MLX installation guidance to reference additional setup instructions.
  * Clarified and reinforced installation reliability by pinning compatible MLX package versions during setup.
* **Tests**
  * Added a Darwin/Apple Silicon MLX smoke test to verify key MLX, audio (TTS/STT), and language (load/generate) entry points import and run.

### DIFF
--- a/backend/requirements-mlx.txt
+++ b/backend/requirements-mlx.txt
@@ -16,7 +16,8 @@ miniaudio>=1.59
 # mlx_audio.stt.load) works fine on transformers 4.57.x in practice.
 #
 # Install it via `pip install --no-deps mlx-audio==0.4.1` after this file
-# (see .github/workflows/release.yml). Most other mlx-audio runtime deps
+# (see .github/workflows/release.yml and the setup-python recipe in the
+# justfile). Most other mlx-audio runtime deps
 # (huggingface_hub, librosa, mlx-lm, numba, numpy, protobuf, pyloudnorm,
 # sounddevice, tqdm) are already in requirements.txt or pulled in by
 # other engines.
--- a/backend/tests/test_mlx_smoke.py
+++ b/backend/tests/test_mlx_smoke.py
@@ -1,0 +1,55 @@
+"""
+Smoke test for the MLX backend dependencies on Apple Silicon.
+
+Guards the `--no-deps` install of mlx-audio/mlx-lm done by `just setup-python`
+and release.yml: those packages skip their declared dependencies (transformers
+>=5.x conflict), so a missing transitive dep only surfaces at import time.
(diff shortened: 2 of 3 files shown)


flowbrix:
Addressed the CodeRabbit nitpick: added `backend/tests/test_mlx_smoke.py` (bb0e990), a pytest-discoverable smoke test that runs under `just test`. It is skipped off Apple Silicon and covers the exact entry points the backend uses — `mlx_audio.tts.load`, `mlx_audio.stt.load` (which also exercises the `miniaudio` dep from #505), `mlx_lm.load`/`generate`, plus a basic `mlx.core` op. Verified locally on Apple Silicon: 4 passed.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #896, opened 2026-07-14
fix(mlx): run all MLX operations on a single dedicated thread — fixes "There is no Stream(gpu, N) in current thread"

### LINKED ISSUE #675: MLX thread error: "There is no Stream(gpu) in current thread" on HTTP REST and MCP transcribe
## Environment
- macOS 26.5, Apple Silicon (M4)
- Voicebox v0.5.0 (app + server)

## Bug summary
Voicebox HTTP REST API (`POST /speak`) and MCP `voicebox.transcribe` both fail with MLX thread-context errors. The MCP `voicebox.speak` tool works fine — the issue is specific to HTTP handlers using MLX in async/FastMCP context.

### 1. HTTP REST `/speak`
```bash
curl -s http://127.0.0.1:17493/speak \
  -X POST \
  -H "X-Voicebox-Client-Id: test" \
  -d "{\"text\":\"hello\",\"profile\":\"<cloned_voice>\"}"
```
→ SSE stream shows `status: "generating"` for ~30s then `status: "failed"` with:
```
There is no Stream(gpu, 1) in current thread.
```

### 2. MCP `voicebox.transcribe`
Call via any MCP client with an audio file:
```
voicebox.transcribe(audio_path="/path/to/audio.wav", language="zh")
```
→ Same error:
```
There is no Stream(gpu, 4) in current thread.
```

## Expected behavior
Both the HTTP speak endpoint and MCP transcribe should work on Apple Silicon with MLX backend, same as the MCP speak tool does.

## Additional context
- This breaks any non-MCP integration (custom TTS providers, shell scripting, etc.)
- The MCP `voicebox.speak` works, suggesting the problem is in how FastMCP HTTP handlers set up the MLX thread context vs the MCP tool execution path

### LINKED ISSUE #699: Qwen TTS generation fails 100% with `Stream(gpu, 1) not in current thread` on Apple Silicon (v0.5.0)

flowbrix:
## Symptom

On macOS (Apple Silicon, MLX backend), Qwen3-TTS voice-clone (ICL) generation fails with HTTP 500 on every request:

```
RuntimeError: There is no Stream(gpu, 1) in current thread.
```

The same error hits the MLX STT path (`voicebox.transcribe` via HTTP/MCP), with varying stream numbers (`Stream(gpu, 4)`, …).

Fixes #675
Fixes #699

## Root cause

`mlx_lm` creates a **thread-local** GPU stream at module import time — [`mlx_lm/generate.py:226`](https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/generate.py) in mlx-lm 0.31.x:

```python
generation_stream = mx.new_thread_local_stream(mx.default_device())
```

Any MLX array or op that captures that stream can only be evaluated on the thread that created it. `backend/backends/mlx_backend.py` dispatched model loading and generation through `asyncio.to_thread`, which uses a **rotating** worker pool — so the import, the load, and each generation could land on different threads. As soon as a generation ran on a thread other than the importing one, MLX raised `There is no Stream(gpu, N) in current thread` and the request failed. (`mlx_audio` has module-level streams on its paths too, so the STT backend is affected the same way.)

This is why behavior looked inconsistent across entry points (#675): whichever call path happened to reuse the importing thread worked; everything else failed.

## Fix

Route all MLX work — module imports, model loading, generation, transcription — through one shared single-worker `ThreadPoolExecutor` (`_MLX_EXECUTOR` + `_run_on_mlx_thread` helper), replacing the four `asyncio.to_thread` call sites in `mlx_backend.py`. Everything MLX now lives on the same dedicated thread for the lifetime of the process.

The serialization is free in practice: MLX is single-GPU and generations were already serialized by the backend.

## Verification

On an M3 Max (macOS 26, mlx-lm 0.31.3):

- Before: Qwen3-TTS 0.6B cloned-voice generation via `/generate/stream` → HTTP 500 (`There is no Stream(gpu, 1) in current thread`), 100% reproducible.
- After: same request succeeds; warm RTF ~0.9. Repeated/concurrent requests stay stable (no thread-affinity dependence on which handler thread serves the request).
- Kokoro TTS and Whisper STT paths re-tested, unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MLX-based speech-to-text and text-to-speech reliability by running model loading and inference through a dedicated execution thread.
  * Reduced the risk of concurrency-related issues during audio processing.

### DIFF
--- a/backend/backends/mlx_backend.py
+++ b/backend/backends/mlx_backend.py
@@ -6,9 +6,25 @@ from typing import Optional, List, Tuple
 import asyncio
 import logging
 import numpy as np
+from concurrent.futures import ThreadPoolExecutor
 from pathlib import Path
 
 logger = logging.getLogger(__name__)
+
+# All MLX work (module imports, model loading, generation) must happen on ONE
+# dedicated thread: mlx_lm creates a thread-local GPU stream at import time,
+# and MLX arrays/ops that captured it fail with "There is no Stream(gpu, N)
+# in current thread" when later evaluated from a different thread — which is
+# exactly what asyncio.to_thread's rotating pool does. Observed on the
+# Qwen3-TTS voice-clone (ICL) path; a single-worker executor fixes it for
+# every MLX backend at once (MLX is single-GPU, so this serialization costs
+# nothing in practice).
+_MLX_EXECUTOR = ThreadPoolExecutor(max_workers=1, thread_name_prefix="mlx")
+
+
+async def _run_on_mlx_thread(fn, *args):
+    """Run a blocking MLX operation on the dedicated MLX thread."""
+    return await asyncio.get_running_loop().run_in_executor(_MLX_EXECUTOR, fn, *args)
 
 # PATCH: Import and apply offline patch BEFORE any huggingface_hub usage
 # This prevents mlx_audio from making network requests when models are cached
@@ -82,7 +98,7 @@ class MLXTTSBackend:
             self.unload_model()
 
         # Run blocking load in thread pool
-        await asyncio.to_thread(self._load_model_sync, model_size)
+        await _run_on_mlx_thread(self._load_model_sync, model_size)
 
     # Alias for compatibility
     load_model = load_model_async
@@ -259,7 +275,7 @@ class MLXTTSBackend:
             return audio, sample_rate
 
         # Run blocking inference in thread pool
-        audio, sample_rate = await asyncio.to_thread(_generate_sync)
+        audio, sample_rate = await _run_on_mlx_thread(_generate_sync)
 
         return audio, sample_rate
 
@@ -293,7 +309,7 @@ class MLXSTTBackend:
             return
 
         # Run blocking load in thread pool
-        await asyncio.to_thread(self._load_model_sync, model_size)
+        await _run_on_mlx_thread(self._load_model_sync, model_size)
 
     # Alias for compatibility
     load_model = load_model_async
@@ -364,4 +380,4 @@ class MLXSTTBackend:
                 return str(result).strip()
 
         # Run blocking transcription in thread pool
-        return await asyncio.to_thread(_transcribe_sync)
+        return await _run_on_mlx_thread(_transcribe_sync)


leighton-tidwell:
Confirming we hit this on **Voicebox 0.5.0**, macOS 26.5.1, Apple Silicon. `RuntimeError: There is no Stream(gpu, 1) in current thread.` on the MLX path — your root-cause writeup (thread-local `generation_stream` created at import in mlx-lm 0.31.x vs. inference on a threadpool) matches our traces exactly. Reproduced with `mlx-community/Qwen3-4B-4bit`. Running all MLX ops on one dedicated thread is the correct fix — would love this merged.

jamiepine:
You nailed the root cause — Metal streams are thread-local, so MLX ops hopping executor threads is exactly what produces "There is no Stream(gpu, N) in current thread". We landed the same fix on `0.6-dictation-fixes` as `backend/services/mlx_thread.py`: a single-worker executor that all MLX work routes through, with load+infer submitted as one job so they can never split across threads. Closing since that supersedes this, but the diagnosis here was spot on. Ships with 0.6.

PR #908, opened 2026-07-18
Fix model cache permission error on startup

zperzendetta:
When the app starts it runs as the voicebox user, but the model cache directory ends up owned by root, so the first model download fails with a permission error.

This happens because the cache is a named volume mounted at a path that does not exist in the image, so it gets created owned by root. The fix creates and owns the directory in the entrypoint before the app user takes over, which repairs it on every start.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Hugging Face cache permissions so the application can write to the cache when using newly created volumes.
  * Improved startup reliability when running the service with restricted privileges.

### DIFF
--- a/scripts/rocm-entrypoint.sh
+++ b/scripts/rocm-entrypoint.sh
@@ -12,4 +12,11 @@ for dev in /dev/kfd /dev/dri/render*; do
     }
     usermod -aG "$grp" voicebox
 done
+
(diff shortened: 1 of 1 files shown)


jamiepine:
Thanks — the cache permission error is real. We're consolidating this into #945, which fixes the same chown issue alongside the Docker torch pin, so closing this one in its favor to keep the Docker changes in one place.

Issue #836, opened 2026-07-01
macOS 26.5.1: Push-to-Talk does not work in native fullscreen apps (recording only starts after leaving fullscreen)


hussaintrawadi:
### Environment

- VoiceBox: v0.5.0
- macOS: 26.5.1 (25F80)
- Device: MacBook Air M4
- Memory: 24 GB Unified Memory

### Description

Dictation works perfectly in normal desktop windows.

However, it completely fails inside native macOS fullscreen applications (Chrome, Whatsapp, etc.).

The behavior is unusual because the hotkey itself appears to be registered, but VoiceBox never begins recording while the app is in fullscreen.

### Steps to Reproduce

1. Install VoiceBox v0.5.0.
2. Grant all requested permissions (Accessibility, Input Monitoring, Microphone).
3. Download Whisper Large v3 Turbo and Qwen 0.6B.
4. Configure Push-to-Talk.
5. Open Google Chrome.
6. Enter native macOS fullscreen (green traffic-light button).
7. Focus any text field.
8. Hold the Push-to-Talk shortcut.

### Expected Behavior

- VoiceBox capsule appears.
- Recording begins immediately.
- Releasing the shortcut transcribes the audio.
- Transcript is inserted into the focused text field.

### Actual Behavior

- Nothing happens while the application remains in fullscreen.
- The VoiceBox capsule never appears.
- Recording does NOT start.
- As soon as I leave fullscreen and return to the desktop, VoiceBox immediately begins recording (as if the original key press had been delayed).
- Recording continues until I press the shortcut again.
- The transcript is generated successfully.
- If a text field is still focused, it pastes there. Otherwise it is only saved inside VoiceBox.

### Additional Information

- Works perfectly on the desktop.
- Works in normal (non-fullscreen) windows.
- Whisper transcription itself works correctly.
- Qwen refinement works correctly.
- Manual model installation through the Hugging Face cache also works correctly.
- This appears to be specifically related to native macOS fullscreen Spaces.

### Things I Already Tried

- Reinstalled models.
- Restarted VoiceBox.
- Changed Push-to-Talk shortcuts.
- Confirmed models are detected.
- Verified permissions.
- Issue still reproduces consistently.

This feels like the fullscreen window never receives or processes the recording trigger until VoiceBox becomes visible again outside of the fullscreen Space.

PR #910, opened 2026-07-18
fix(dictation): warm the mic so push-to-talk doesn't clip the first words

mcox5:
## Problem

On push-to-talk dictation the **first ~1 second of speech — often the first few words — is clipped** on every capture. Steps to reproduce (macOS, Apple Silicon):

1. Configure a push-to-talk shortcut and grant Microphone / Accessibility / Input Monitoring.
2. Focus a text field, hold the shortcut and **start speaking immediately**.
3. Release. The transcript consistently drops the opening words.

## Root cause

`useAudioRecording.startRecording()` opens a **fresh** `getUserMedia` stream on every recording and stops its tracks again on every stop:

```ts
const stream = await navigator.mediaDevices.getUserMedia({ audio: {...} }); // slow on macOS
...
mediaRecorder.start(); // only starts capturing AFTER the await resolves
```

On macOS opening the input device takes several hundred ms — up to ~1s cold — and `MediaRecorder` only begins capturing once the promise resolves. Everything spoken in that window is lost, and because the stream is torn down after each capture, **every** dictation pays the cost, not just the first.

This is distinct from the fullscreen bug (#836 / #848, where `getUserMedia` never resolves at all in a native-fullscreen Space) and from #843 (raw-audio constraints / waveform stream) — neither keeps the dictation mic warm.

## Changes

- **`useAudioRecording`** — new opt-in `keepWarm` option. When set, the capture `MediaStream` is kept open between recordings and reused, so the hot path (chord-down → record) never awaits `getUserMedia` and `MediaRecorder` captures from the first frame. Adds a `prewarm()` that pre-opens the device, and a dead-stream guard that re-acquires if the tracks have ended (device unplugged / switched). The warm stream self-releases after `WARM_IDLE_RELEASE_MS` (60s) of inactivity so the OS mic indicator clears when the user isn't dictating.
- **`useCaptureRecordingSession`** — new `keepMicWarm` option, threaded through to `useAudioRecording`, plus a mount effect that calls `prewarm()` so even the first dictation is warm.
- **`DictateWindow`** — the always-mounted floating pill opts in with `keepMicWarm`. Warming lives here (not the main window) so the mic is only held open by the surface that actually services the global shortcut.

Behavior is **unchanged** for the voice-clone sample recorders (`ProfileForm`, `SampleUpload`) and the main-window Dictate button — they don't pass `keepWarm`/`keepMicWarm`, so they still open once and release the device immediately.

## Trade-off / notes

- With `keepWarm` the mic is held open (indicator lit) during active dictation and for up to `WARM_IDLE_RELEASE_MS` after — the price of zero clipping. Happy to make the timeout configurable or gate warming behind a setting if you'd prefer.
- The one case still not covered is the *very first* dictation after the idle window elapses (it re-warms lazily). A clean follow-up would be to emit a lightweight "chord armed" event from the Rust hotkey monitor on modifier-down and call `prewarm()` on it, fully eliminating clipping without holding the device open at all.

## Verification

- `bunx tsc -p app/tsconfig.json --noEmit` — passes.
- `biome lint` on the changed files — clean (no new warnings; the pre-existing `useExhaustiveDependencies` warnings in `DictateWindow` are untouched).
- Manually: on macOS 26.5.2 / Apple Silicon, held the shortcut and spoke immediately — opening words are now captured; back-to-back dictations start instantly.

## Summary by CodeRabbit

* **New Features**
  * Improved dictation startup by preparing the microphone in advance, helping prevent clipped words at the beginning of recordings.
  * Kept the microphone ready between recordings for faster subsequent dictation.
  * Automatically releases the microphone after inactivity to conserve device resources.
* **Bug Fixes**
  * Improved microphone handling when recording is canceled or stopped.
  * Gracefully handles unavailable microphone permissions or devices during pre-activation.

### DIFF
--- a/app/src/components/DictateWindow/DictateWindow.tsx
+++ b/app/src/components/DictateWindow/DictateWindow.tsx
@@ -42,6 +42,10 @@ export function DictateWindow() {
   const focusRef = useRef<FocusSnapshot | null>(null);
 
   const session = useCaptureRecordingSession({
+    // This floating window is always mounted, so keeping the mic warm here
+    // removes the first-words clipping on push-to-talk without holding the
+    // device open from the main app window.
+    keepMicWarm: true,
     onFinalText: async (text, _capture, allowAutoPaste) => {
       const focus = focusRef.current;
       // Consume-once: a second chord before this fires would overwrite
@@ -141,9 +145,7 @@ export function DictateWindow() {
     audio.onplaying = () => {
       emit('dictate:show').catch(() => {});
       setSpeaking((prev) =>
-        prev && prev.generationId === generationId
-          ? { ...prev, startedAt: Date.now() }
-          : prev,
+        prev && prev.generationId === generationId ? { ...prev, startedAt: Date.now() } : prev,
       );
       setSpeakElapsed(0);
     };
--- a/app/src/lib/hooks/useAudioRecording.ts
+++ b/app/src/lib/hooks/useAudioRecording.ts
@@ -5,11 +5,47 @@ import { convertToWav } from '@/lib/utils/audio';
 interface UseAudioRecordingOptions {
   maxDurationSeconds?: number;
   onRecordingComplete?: (blob: Blob, duration?: number) => void;
+  /**
+   * Keep the microphone ``MediaStream`` open between recordings instead of
+   * tearing it down on every stop. This is what removes the "first words get
+   * clipped" problem on push-to-talk dictation: ``getUserMedia`` on macOS can
+   * take several hundred ms — up to a second cold — to hand back a stream, and
+   * ``MediaRecorder`` only starts capturing *after* it resolves, so everything
+   * spoken in that window is lost. With a warm stream already open, the next
+   * ``startRecording`` skips ``getUserMedia`` entirely and ``MediaRecorder``
+   * captures from the first frame.
+   *
+   * Off by default so the voice-clone sample recorders (which record once and
+   * should release the device immediately) keep their existing behavior; only
+   * the dictation session opts in. When on, the stream is released after
+   * ``WARM_IDLE_RELEASE_MS`` of inactivity so the OS mic indicator doesn't stay
+   * lit forever when the user isn't dictating.
+   */
+  keepWarm?: boolean;
 }
+
+// Audio constraints for capture. Kept identical to the previous inline value so
+// this change is purely about *when* the stream is opened, not *how*.
+const AUDIO_CONSTRAINTS: MediaTrackConstraints = {
+  echoCancellation: true,
+  noiseSuppression: true,
+  autoGainControl: true,
+};
+
+// How long a warm stream lingers after the last recording before it's released.
+// Long enough that back-to-back dictations reuse the same warm stream (no
+// clipping), short enough that the mic indicator clears soon after the user
+// stops. A follow-up could re-warm just-in-time on a chord-arming event from
+// Rust to also cover the first dictation after this window elapses.
+const WARM_IDLE_RELEASE_MS = 60_000;
+
+const streamHasLiveAudio = (stream: MediaStream | null): stream is MediaStream =>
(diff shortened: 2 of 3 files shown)


leighton-tidwell:
Confirming this on **Voicebox 0.5.0** / macOS 26.5.1 / Apple Silicon. The first ~1s (first few words) of every push-to-talk capture is clipped, exactly as described — cold `getUserMedia` open + `MediaRecorder` only starting after the promise resolves, with the stream torn down each capture so *every* dictation pays the cost. Wispr Flow doesn't do this because it keeps the mic warm. Warming the stream is the right fix — would love to see this merged.

dknoodle:
Confirming this on **Windows 11 / WebView2 / CUDA** with measurements — same ~1 s clip, every capture, and I can add device-exonerating data that supports keeping the fix at the `getUserMedia`/stream layer where this PR puts it.

## Environment

- Windows 11 Pro (build 26200), i9-11900K, 128 GB RAM, RTX 4090 (driver 32.0.15.8157)
- Voicebox 0.5.0 desktop, CUDA backend `cu128-v1`, WebView2 Runtime 150.0.4078.65
- Mic: Shure MVX2U (USB audio interface), 48 kHz, push-to-talk chord dictation

## Finding 1 — every capture carries exactly 1.0 s of dead audio at the head

RMS analysis (100 ms windows) of 14 consecutive dictation WAVs from `%APPDATA%\sh.voicebox.app\captures\`: **every file starts with 1.0 s of digital near-silence** before speech, metronomically consistent across all 14. Example profiles (`.` = digital silence, `x`/`#` = speech):

```
4.1s  [:.........xxxxxxx####xxxxx#xxxxxxxxxxx::::]   "Let's test transcription and see what happens."
1.9s  [:.........xxxxxxxxx]                           "That is."   (opening words lost)
1.0s  [:..........]                                   ALL SILENT → transcribed as "you"
```

Short utterances land entirely inside the dead window: the file is ~1 s of silence and Whisper hallucinates its classic `"you"` on empty audio — so users see phantom "you" transcripts, which may explain some odd-transcript reports too.

On Windows the lost second is *recorded as zeros* (the stream resolves, then delivers zero-fill while Chromium's audio pipeline spins up) rather than absent, but the user-facing effect is identical to your macOS trace: everything spoken in the first second of a push-to-talk is gone, on **every** capture, exactly as your root-cause section describes (fresh `getUserMedia` per recording).

## Finding 2 — the device/OS stack is not a factor (supports the warm-stream approach)

Raw WASAPI capture on the same mic via PortAudio/sounddevice, bypassing WebView2:

| Test | Stream-open time | Leading digital silence |
|---|---|---|
| Cold #1 | 45 ms | **0.0 s** |
| Warm #2 (immediately after) | 18 ms | **0.0 s** |
| Cold #3 (after 10 s idle) | 18 ms | **0.0 s** |

Live room audio from the first 100 ms window in all three runs; USB power management on the device's audio interface node is not in play. So the entire 1.0 s is inside the WebView2 `getUserMedia` → `MediaRecorder` path — there's nothing to fix at the OS/driver layer, and a warm `MediaStream` (this PR) is the right layer for the fix on Windows as well as macOS.

## Minor related observation

Most captures also show **0.0 s trailing silence** — speech runs flush to the end of the file, so releasing the chord mid-final-word clips the tail. If the `MediaRecorder` stop path doesn't already flush a final timeslice after key-up, a small stop-delay might be worth considering here or in a follow-up.

Happy to test a Windows build of this branch if useful — this is currently the biggest UX gap for our team's dictation use, and the warm-stream approach matches everything we measured.

jamiepine:
Thanks — the first-words clipping is real and you diagnosed it correctly. This got fixed on `0.6-dictation-fixes` as part of a broader mic-lifecycle rework (5787e36) that makes the whole capture path race-safe rather than just pre-warming, so closing this one in favor of that. It ships with 0.6.

PR #946, opened 2026-07-22
fix: .gitignore tail is UTF-16 encoded, so its patterns never match

thiagomargatho:
## Summary

The final block of `.gitignore` — `# Local AI assistant workspace settings` / `.claude/settings.local.json` — is encoded as **UTF-16LE with NUL bytes** (visible with `tail .gitignore | cat -A`; likely appended by a Windows/PowerShell `>>` redirect, which defaults to UTF-16). Git reads `.gitignore` byte-wise, so:

- `.claude/settings.local.json` is currently **not ignored** at all;
- any pattern appended after that block inherits a stray NUL and silently fails to match too.

This PR re-encodes the block as plain ASCII, and adds `output/` — which `docker-compose.yml` bind-mounts to `/app/data/generations` — so generated audio no longer shows up as untracked in Compose deployments.

## Verification

```
$ git check-ignore output/ .claude/settings.local.json
output/
.claude/settings.local.json
```

Before the fix, both commands exit 1 (no match). `git diff` shows the file as binary precisely because of the NUL bytes being removed.


jamiepine:
Great find — and funny timing: we independently caught the UTF-16 tail the same week while auditing the repo, and fixed it on `0.6-dictation-fixes` (tail re-encoded to UTF-8, CRLF normalized to LF throughout). So the underlying bug is dead once 0.6 lands.

One thing your PR had that ours didn't: the `output/` ignore, which the Docker compose bind-mounts. Cherry-picked that with credit to you in 59782e9. Thanks!

PR #940, opened 2026-07-21
feat: BYO-LLM conversational voice mode — OpenAI-compatible endpoint + voice agent loop

andreolf:
## Summary

Turns Voicebox from a TTS studio into a local voice agent: connect any OpenAI-compatible LLM (Ollama, LM Studio, OpenAI, Anthropic, etc.), speak or type to it, and hear replies in a cloned voice — running entirely on your machine.

This is the text-LLM bridge that the roadmap lists as the stepping stone before end-to-end speech LLMs (Moshi, GLM-4-Voice). It reuses the entire existing generation pipeline unchanged.

### What's new

**Backend**
- `ConversationSettings` ORM singleton (same pattern as `CaptureSettings`) — `enabled`, `llm_endpoint`, `llm_api_key`, `llm_model`, `system_prompt_prefix`
- `backend/backends/openai_compat_backend.py` — `OpenAICompatLLMBackend` via `httpx` (already a dependency); implements both `generate()` and `generate_with_history()` for multi-turn chat
- `POST /conversation/turn` — validates settings → calls external LLM with the profile's personality as system prompt + conversation history → enqueues TTS via the existing background job pipeline → returns `assistant_text` + `generation_id` immediately (non-blocking, same pattern as `/generate`)
- `GET /PUT /settings/conversation` — settings CRUD, same pattern as captures/generation settings

**Frontend**
- `ConversationPanel.tsx` — chat UI with user/assistant bubbles, Enter-to-send, polls SSE endpoint and auto-plays audio via the existing audio player store
- `ConversationPage.tsx` — settings page: enable toggle, endpoint URL, API key (password field), model name, system-prompt textarea, provider compatibility sidebar (Ollama · LM Studio · OpenAI · any compatible API)
- "Conversation" entry added to settings nav + TanStack Router
- "Voice Chat" tab in VoiceInspector — visible only when conversation mode is enabled

## Compatibility

Works out of the box with any server that speaks `/v1/chat/completions`:
- **Ollama** — `http://localhost:11434`, no API key
- **LM Studio** — `http://localhost:1234`, no API key  
- **OpenAI** — `https://api.openai.com`, API key required
- **Anthropic** — via compatible proxy

## Test plan

- [ ] Enable conversation mode in Settings → Conversation
- [ ] Set endpoint to a running Ollama instance (`ollama run llama3`)
- [ ] Open a voice profile → Voice Chat tab → send a message
- [ ] Verify reply text appears and audio plays in the cloned voice
[...]

PR #947, opened 2026-07-22
Add custom OpenAI-compatible LLM endpoint for refinement and personality

### LINKED ISSUE #857: Unlocking Voicebox for everyone: Optional cloud fallbacks to bypass local constraints
While Voicebox's local-first design is highly valued, local setup (CUDA/MPS, MLX, slow gen, and massive size downloads) poses a significant barrier for many users. An analysis of the ~70+ open issues reveals that **roughly 40% to 45% of them** (related to hardware setup, crashes, network blocks, or unsupported languages) could be bypassed or resolved by introducing an optional **Cloud Fallback** mode.

I propose adding support for external APIs as alternative engines:
* **Open Router:** For TTS, STT, LLM and more through 1 api.
* **TTS:** ElevenLabs, Cartesia, or OpenAI.
* **STT:** OpenAI Whisper.
* **LLM (Refinement/Personality):** OpenRouter, OpenAI, or Anthropic.

This would keep local execution as the default, but offer a cloud option that expands accessibility to older hardware and drastically reduces the maintenance overhead of troubleshooting local environment issues.

ext-sakamoro:
Closes #857 (partial — LLM cloud fallback path).

Adds a settings-driven way to swap the built-in Qwen3 LLM for any endpoint that speaks OpenAI's `/v1/chat/completions` protocol: OpenAI itself, OpenRouter, Anthropic-via-proxy, llama.cpp server, vLLM, LM Studio, LocalAI, Ollama's OpenAI shim, or a self-hosted engine. An empty endpoint field keeps refinement and personality-speak running on Qwen exactly like before.

## Design

- New `OpenAICompatLLMBackend` implementing the existing `LLMBackend` protocol. Uses `httpx.AsyncClient`; non-streaming for this PR (a streaming variant lands in #948).
- Module-level `set_llm_config()` + a `get_llm_backend()` dispatch that reroutes to the remote backend when the endpoint is configured. Config changes drop the cached backend so URL / model swaps take effect without a restart.
- Three new `capture_settings` columns (`custom_llm_endpoint`, `custom_llm_model`, `custom_llm_api_key`) with an idempotent column-add migration matching the existing `_migrate_capture_settings` style, and matching `CaptureSettingsResponse` / `CaptureSettingsUpdate` fields.
- Startup bootstrap seeds `set_llm_config()` from the persisted row before the first LLM call; every `PUT /settings/captures` re-syncs.
- `/llm/generate` route skips the Qwen size validation and download-progress branch when the custom endpoint is active — the remote server owns model selection.
- Frontend adds a **Custom LLM endpoint (advanced)** section under Refinement in Captures settings with three inputs (endpoint URL, model name, API key) that persist on blur. A status footnote reports whether the built-in Qwen3 or the remote endpoint is currently active. en / ja translations included; other locales fall back to English until a native pass lands.

## Testing

- `backend/tests/test_openai_compat_integration.py` — real HTTP round-trip (no mocks) against a caller-configured endpoint (`VOICEBOX_TEST_OPENAI_COMPAT_URL` / `_MODEL` env vars), covering protocol compliance, few-shot history propagation, dispatch, cache invalidation, and Qwen fallback. Skips automatically when the endpoint isn't reachable so `pytest` on a fresh clone stays green.
- Manually verified end-to-end against a local llama.cpp-style OpenAI-compatible server: `PUT /settings/captures` persists the endpoint, `POST /llm/generate` reaches the remote model, coherent responses come back, and clearing the endpoint reverts to the on-device Qwen path.

## Relationship to #940

@andreolf's #940 (BYO-LLM conversational voice mode) targets a different UX — a new voice-chat surface with `ConversationSettings` + `ConversationPanel` — but lands on the same shared backend file (`backend/backends/openai_compat_backend.py` with an `OpenAICompatLLMBackend`). Feature scopes are cleanly independent (different DB tables, different frontend surfaces) but the backend class needs to be de-duplicated when both PRs land. Happy to rebase this one onto the shared class if #940 lands first, and posted the same note on that PR so we can coordinate.

## Breaking changes / migration

None. Existing installs get three new nullable columns and default to `null` — every existing behaviour test still passes on the on-device Qwen path.

## Follow-up

A streaming LLM → TTS pipeline that layers on top of this endpoint infrastructure lands in #948 (progressive audio playback, ~7 s first-audio-latency improvement measured against a mock TTS). Kept separate so this one stays reviewable in isolation.

## Summary by CodeRabbit

* **New Features**
  * Added advanced custom OpenAI-compatible LLM configuration for capture refinement (endpoint URL, model, and a saved “API key configured” indicator).
  * Refinement/personality rewriting can now run remotely via the custom endpoint while transcription stays local.
* **Bug Fixes**
  * Improved switching between remote and built-in refinement by initializing and syncing the active LLM backend from saved settings.
  * Updated remote routing so model-size handling matches the selected backend.
* **Documentation**
  * Added English and Japanese translations for the new custom-endpoint UI and updated “remote vs local” wording.
* **Tests**
  * Added live integration coverage for the OpenAI-compatible backend.

### DIFF
--- a/app/src/components/ServerTab/CapturesPage.tsx
+++ b/app/src/components/ServerTab/CapturesPage.tsx
@@ -143,7 +143,10 @@ export function CapturesPage() {
   const toggleToTalkKeys = settings?.chord_toggle_to_talk_keys ?? defaultChordKeys('toggle');
   const customLlmEndpointSaved = settings?.custom_llm_endpoint ?? '';
   const customLlmModelSaved = settings?.custom_llm_model ?? '';
-  const customLlmApiKeySaved = settings?.custom_llm_api_key ?? '';
+  // The server no longer returns the API key value; it reports whether one
+  // is stored via ``custom_llm_api_key_configured``. Anything sitting in
+  // React state below is a freshly-typed key that hasn't been sent yet.
+  const customLlmApiKeyConfigured = Boolean(settings?.custom_llm_api_key_configured);
   const customLlmActive = Boolean(customLlmEndpointSaved && customLlmModelSaved);
 
   const [chordEditor, setChordEditor] = useState<'push' | 'toggle' | null>(null);
@@ -153,7 +156,10 @@ export function CapturesPage() {
   // re-render on every keystroke while the user types out a URL or token.
   const [customLlmEndpointDraft, setCustomLlmEndpointDraft] = useState(customLlmEndpointSaved);
   const [customLlmModelDraft, setCustomLlmModelDraft] = useState(customLlmModelSaved);
-  const [customLlmApiKeyDraft, setCustomLlmApiKeyDraft] = useState(customLlmApiKeySaved);
+  // Write-only field: the input is blank on load even when a key is stored,
+  // and clears itself after a successful save so the raw value never lingers
+  // in React state where a browser cache or dev-tools inspection could grab it.
+  const [customLlmApiKeyDraft, setCustomLlmApiKeyDraft] = useState('');
 
   useEffect(() => {
     setCustomLlmEndpointDraft(customLlmEndpointSaved);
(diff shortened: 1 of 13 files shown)


ext-sakamoro:
Note for anyone rereading: the CodeRabbit findings on this PR (privacy statement, `content.strip` type guard, API key write-only) were already addressed in `f009322`. Pushed one more fix (`c7a9d4f`) for a related trim bug in `commitCustomLlmApiKey` that the AI Code Review Agent caught on the downstream PR #948 — the function was introduced in `bb866fa` on this branch, so the fix belongs here too.

PR #948, opened 2026-07-22
Add streaming LLM→TTS pipeline with SSE audio to client

ext-sakamoro:
Depends on #947 (custom OpenAI-compat endpoint). This PR stacks on top of it — every commit from #947 is included, with the streaming work in the last eight.

Layers a streaming path onto the endpoint infrastructure from #947 so users pointed at an SSE-capable OpenAI-compat endpoint (llama.cpp server, vLLM, LM Studio, LocalAI, OpenAI, or a self-hosted engine like [ALICE-LLM](https://github.com/ext-sakamoro/ALICE-LLM)) hear the first sentence before the LLM has finished generating.

### Backend

- `LLMBackend.generate_stream()` + `supports_streaming()` added to the protocol. `OpenAICompatLLMBackend` implements a real SSE consumer via `httpx.AsyncClient.stream()` — yields `delta.content` strings as they arrive, honours the `[DONE]` sentinel. The built-in Qwen3 backends declare no streaming and fall back to a single-yield wrap of `generate()`.
- `chunked_tts` gains three async helpers: `stream_sentences` (LLM chunk iter → sentence iter), `generate_streaming_from_sentences` (fires per-sentence TTS as `asyncio.create_task`, yields ordered `(audio, sr, sentence)` 3-tuples), and `generate_streaming_chunked` end-to-end.
- `services.personality.rewrite_as_profile_stream` mirrors `rewrite_as_profile` but yields per-token deltas so the streaming TTS pipeline can start work before the LLM has finished restating the message.
- `services.generation.run_generation` grows a `personality_prompt` parameter. When set alongside a streaming-capable LLM backend, the background task runs the streaming pipeline instead of the sequential text → chunked TTS path. Backends without streaming ignore the parameter and take the existing route.
- **New \`POST /speak/stream\` route** returns \`text/event-stream\` — same profile / engine / personality resolution as \`/speak\`, but the response body is a sequence of SSE frames (\`meta\` → \`audio\` per sentence → \`complete\` → \`[DONE]\`). Audio is base64-encoded PCM float32 in a JSON payload; final concatenated audio still lands on disk and gets a generations-table row so History keeps working.

### Frontend

- \`apiClient.streamSpeak(params, onEvent, signal)\` opens the SSE stream with \`fetch\` + \`ReadableStream\`, parses SSE \`\\n\\n\` frame boundaries + repeated \`data:\` lines, dispatches events to a callback, and rejects with \`AbortError\` on user cancellation.
- \`useStreamingSpeak\` hook — state machine (idle → connecting → streaming → complete | error | aborted), single \`AudioContext\` opened lazily inside a user gesture, gap-less \`nextStartTime\` cursor so consecutive sentences abut with zero gap, \`playingIndex\` that only moves forward, unmount cleanup, \`AbortController\` cancellation.
- \`FloatingGenerateBox\` renders a secondary **Speak with streaming (β)** button whenever the user has both wired a custom endpoint AND selected a profile with a personality prompt set — the two conditions where the LLM+TTS overlap actually buys any latency. Otherwise the button stays hidden and the box looks exactly like it did before. Icon flips from \`AudioLines\` to \`Square\` while streaming; click aborts.

## Measured latency

Against a real streaming source (an OpenAI-compat SSE endpoint) with a mock TTS backend rendering each sentence in 500 ms, a two-sentence weather-forecast prompt reached first audio ~7 s earlier than the sequential LLM→TTS path (29 s vs 36 s wall clock). Real TTS engines (Kokoro, Chatterbox, Qwen CustomVoice) will amplify the total-time win too because their per-sentence latency is closer to the LLM's than the mock is.

## Testing

- **Backend**: module-layer end-to-end drives \`stream_speak_events\` against a real SSE endpoint + MockTTS and verifies the frame sequence (\`meta\` → \`audio × N\` → \`complete\` → \`[DONE]\`), base64 PCM round-trip, error-frame fallback, and generations-row persistence.
- SSE protocol verified via \`curl -N\` — per-token \`delta.content\` chunks arrive incrementally, terminal \`finish_reason\` chunk lands, and the non-streaming path (\`stream=false\`) preserves the existing \`ChatCompletionResponse\` shape.

**Not yet verified in-browser.** The UI parts (Web Audio scheduling, autoplay policy handling, subtitle rendering) follow the standard patterns for progressive audio playback but I don't have a Tauri dev environment set up here to click through the actual button. The (β) suffix on the label reflects that. Happy to iterate on visual issues once someone with the setup can play with it.

## Breaking changes / migration

None. \`POST /speak/stream\` is additive alongside \`POST /speak\`. The streaming path in \`run_generation\` only fires when the LLM backend reports \`supports_streaming=True\` AND \`personality=true\`; the existing sequential path is preserved otherwise, and the streaming button UI only appears when the two settings that make it useful are both on.

## Summary by CodeRabbit

* **New Features**
  * Added streaming “Speak” with start/stop controls and progressive, sentence-by-sentence playback.
  * Added advanced custom LLM (OpenAI-compatible) settings for refinement/personality rewriting, including remote-vs-built-in status and API key replace/clear.
* **Bug Fixes**
  * Improved streaming reliability with proper cancellation/abort behavior and clearer mid-stream error feedback; results are persisted on successful completion.
* **Documentation**
  * Updated English and Japanese UI text for streaming controls and remote refinement descriptions.
* **Tests**
  * Added real OpenAI-compatible backend integration tests that automatically skip when not configured or unreachable.

### DIFF
--- a/app/src/components/Generation/FloatingGenerateBox.tsx
+++ b/app/src/components/Generation/FloatingGenerateBox.tsx
@@ -508,7 +508,11 @@ export function FloatingGenerateBox({
                 </AnimatePresence>
 
                 <AnimatePresence>
-                  {streamingAvailable && (
+                  {/* Keep the button mounted while a stream is active — otherwise
+                      clearing the text field or switching to a profile without
+                      a personality would remove the only in-place way to stop
+                      audio that ``useStreamingSpeak`` has already scheduled. */}
+                  {(streamingAvailable || streamingActive) && (
                     <motion.div
                       initial={{ opacity: 0, scale: 0.8 }}
                       animate={{ opacity: 1, scale: 1 }}
@@ -520,7 +524,12 @@ export function FloatingGenerateBox({
                           type="button"
                           variant="ghost"
                           size="icon"
-                          disabled={!selectedProfileId || !form.watch('text')?.trim()}
+                          // Validation only matters when starting a new stream;
+                          // an active stream must always be abortable.
+                          disabled={
+                            !streamingActive &&
+                            (!selectedProfileId || !form.watch('text')?.trim())
+                          }
                           onClick={() => {
                             if (streamingActive) {
                               streamAbort();
--- a/app/src/components/Generation/FloatingGenerateBox.tsx
+++ b/app/src/components/Generation/FloatingGenerateBox.tsx
@@ -118,6 +118,19 @@ export function FloatingGenerateBox({
   const streamingActive =
     streamState.status === 'connecting' || streamState.status === 'streaming';
 
+  // Surface mid-stream faults through the shared toast — otherwise the
+  // streaming button just flips back to idle and the user has no idea
+  // whether their utterance failed or finished silently.
+  useEffect(() => {
+    if (streamState.status !== 'error') return;
+    toast({
+      title: t('generation.button.streamErrorTitle'),
+      description:
+        streamState.error ?? t('generation.button.streamErrorFallback'),
+      variant: 'destructive',
+    });
+  }, [streamState.status, streamState.error, toast, t]);
+
   // Click away handler to collapse the box
   useEffect(() => {
     function handleClickOutside(event: MouseEvent) {
--- a/app/src/components/Generation/FloatingGenerateBox.tsx
+++ b/app/src/components/Generation/FloatingGenerateBox.tsx
@@ -1,7 +1,7 @@
 import { useMutation, useQuery } from '@tanstack/react-query';
 import { useMatchRoute } from '@tanstack/react-router';
 import { AnimatePresence, motion } from 'framer-motion';
-import { Dices, Loader2, SlidersHorizontal, Sparkles, Wand2 } from 'lucide-react';
+import { AudioLines, Dices, Loader2, SlidersHorizontal, Sparkles, Square, Wand2 } from 'lucide-react';
 import { useEffect, useRef, useState } from 'react';
 import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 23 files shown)


kemo04 (commented):
## AI Code Review (PR Review Agent)

- **[Logic & Correctness · Medium]** — `app/src/components/ServerTab/CapturesPage.tsx` (diff line 186): commitCustomLlmApiKey sends `customLlmApiKeyDraft` instead of the pre-trimmed `trimmed` variable.
- **[Test Coverage · Critical]** — `app/src/components/Generation/FloatingGenerateBox.tsx` (diff line 40): New streaming availability logic based on custom LLM settings and profile personality lacks test coverage.
- **[Test Coverage · Critical]** — `app/src/components/Generation/FloatingGenerateBox.tsx` (diff line 48): Error toast notification effect on stream failure lacks test coverage.

ext-sakamoro:
Addressed the AI Code Review Agent finding (Logic & Correctness, Medium):

`commitCustomLlmApiKey` was computing `trimmed` for the empty-check but then passing the untrimmed `customLlmApiKeyDraft` to `update()`. Bearer tokens pasted from Notion / a code block would carry surrounding whitespace to the remote and get rejected with 401 as a confusing 'auth failed' toast even though the raw key was fine.

Fix: send `trimmed` on line 189 (`45d6987`). The test-coverage findings on `FloatingGenerateBox.tsx` (streaming availability logic + error toast) are worth a follow-up but I'd like to keep this PR focused on the bug fix.

PR #959, opened 2026-07-23
Implement cpal device selection by PulseAudio source name

goat82447-ops:
Added a function to find cpal input devices by PulseAudio source name. Updated audio capture logic to select devices based on detected monitor sources.
AB#https://github.com/jamiepine/voicebox/pull/457.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Linux audio capture device selection for PulseAudio and PipeWire monitor sources by matching detected monitor names to available input devices.
  * Added robust fallback behavior when a monitor can’t be matched, including alternate monitor-style devices or the default input device.
  * Reduced capture failures caused by incorrect system-wide audio source selection.


jamiepine:
Thanks — you correctly identified the `PULSE_SOURCE` `set_var` race, which our own audit flagged the same week. We went with #949 for this one (now merged), which removes the env mutation entirely rather than working around it. Closing as the cluster is resolved, but the diagnosis was right.

Issue #577, opened 2026-04-28
Hope to be able to change the locations of the "Generate Folder" and "Capture Folder"


iambestt:
Currently, the models can be moved to a different location, but only after they have been downloaded. It would be much better if we could download them directly to a specific folder. >
Additionally, I’d like the option to change the default locations for the "Generated" and "Captured" folders. Thanks!

PR #1004, opened 2026-08-08
fix: honor VOICEBOX_DATA_DIR in both the backend and the desktop app

### LINKED ISSUE #533: VOICEBOX_DATA_DIR it doesn't seem to work
I see Default location is the OS-specific app data directory. Override with --data-dir or the VOICEBOX_DATA_DIR environment variable.
so

I configured the VOICEBOX_DATA_DIR environment variable on Windows 11, but the application ignores it. Even after a system restart, it continues to default to the C drive.

version: Voicebox_0.4.5_x64-setup.exe

2026-04-22 22:05:09,662 - __main__ - INFO - Arguments: ['C:\\Users\\lihong\\AppData\\Roaming\\sh.voicebox.app\\backends\\cuda\\voicebox-server-cuda.exe', '--data-dir', 'C:\\Users\\lihong\\AppData\\Roaming\\sh.voicebox.app', '--port', '17493', '--parent-pid', '432']

### LINKED ISSUE #981: VOICEBOX_DATA_DIR is ignored by the desktop app (docs say it overrides the data directory)
## Summary

`backend/README.md` states:

> Default location is the OS-specific app data directory. Override with `--data-dir` or the `VOICEBOX_DATA_DIR` environment variable.

In the **desktop app** the environment variable has no effect: the Tauri shell always passes `--data-dir` explicitly to the sidecar, and the explicit argument wins over the env var.

## Environment

Voicebox v0.5.0, Windows 11 Pro 25H2, MSI install.

## Steps to reproduce

1. Set a user-level environment variable:
```powershell
   [Environment]::SetEnvironmentVariable('VOICEBOX_DATA_DIR','C:\Users\<user>\Audio\Voicebox','User')
```
2. Confirm it propagated, in a **new** shell: `$env:VOICEBOX_DATA_DIR` → prints the new path
3. Create the target directory
4. Fully quit and relaunch Voicebox
5. `curl http://127.0.0.1:17493/health/filesystem`

## Actual

```json
{"healthy":true,"disk_free_mb":327617.7,"disk_total_mb":974485.0,
 "directories":[
  {"path":"C:\\Users\\<user>\\AppData\\Roaming\\sh.voicebox.app\\generations","exists":true,"writable":true,"error":null},
  {"path":"C:\\Users\\<user>\\AppData\\Roaming\\sh.voicebox.app\\captures","exists":true,"writable":true,"error":null},
  {"path":"C:\\Users\\<user>\\AppData\\Roaming\\sh.voicebox.app\\profiles","exists":true,"writable":true,"error":null},
  {"path":"C:\\Users\\<user>\\AppData\\Roaming\\sh.voicebox.app","exists":true,"writable":true,"error":null}]}
```

The target directory stays empty. Server logs show the sidecar being launched with the AppData path passed explicitly:

Arguments: ['\\?\C:\Program Files\Voicebox\voicebox-server.exe', '--data-dir', 'C:\Users\<user>\AppData\Roaming\sh.voicebox.app', '--port', '17493', '--parent-pid', '...']

## Expected

Either the env var is honoured by the desktop app, or the documentation scopes it to standalone backend use.

## Why this matters

The data directory grows without bound: every dictation capture stores raw audio plus transcript, and every generation stores a WAV per version and per take. Users on a small system drive — or who want voice data on a specific volume for backup, encryption or retention reasons — currently have no supported way to relocate it.

There is also no folder picker in Settings: `Generation → Generations folder` and `Captures → Storage` both offer an **Open** button only. The models directory *can* be relocated (`VOICEBOX_MODELS_DIR`, plus an in-app migration flow), which makes the asymmetry more surprising — the relocatable data is the disposable, re-downloadable part, while the irreplaceable user data is pinned to AppData.

Voice recordings are biometric data in several jurisdictions. Being unable to choose where they land is a real constraint for professional use, not just a preference.

## Suggested fixes (any one would resolve it)

1. Have the Tauri shell read `VOICEBOX_DATA_DIR` before computing the path it passes to the sidecar.
2. Add a data-directory setting in Settings → General with a migration flow, mirroring the existing models-folder migration.
3. If neither is planned, update `backend/README.md` to state that the env var applies only when running the backend standalone.

## Workaround (Windows)

A directory junction works and survives normal app updates, since installers do not touch `AppData\Roaming`:

mklink /J "%APPDATA%\sh.voicebox.app" "D:\path\to\voicebox-data"

Verified working on v0.5.0: after relinking, the in-app `Generations folder` path resolves to and displays the junction target, and the backend reads and writes through it without error.

Lvigentini:
`backend/README.md` documents `VOICEBOX_DATA_DIR` alongside `--data-dir` as a way to override the data directory, but it never worked. Two open issues report this: #981 and #533.

## Why it was broken in two places

**Backend.** `config.py` only wired up `VOICEBOX_MODELS_DIR`; the data directory default stayed hardcoded to `./data`.

**Desktop app.** Fixing the backend alone is not enough. The Tauri shell resolved the directory from `app.path().app_data_dir()` unconditionally and passed it to the sidecar as an explicit `--data-dir`, which beats the environment variable on the Python side. Both issue reports show it in the log:

```
Arguments: [...'voicebox-server.exe', '--data-dir',
            'C:\Users\<user>\AppData\Roaming\sh.voicebox.app', ...]
```

So the variable could only ever reach a bare `uvicorn backend.main:app`, never the packaged app.

## Changes

- `backend/config.py` — read `VOICEBOX_DATA_DIR` when computing the default. `--data-dir` still takes precedence, since it calls `set_data_dir()` after import. This is also the only override that reaches `just dev`, which starts uvicorn directly and never parses CLI arguments.
- `tauri/src-tauri/src/main.rs` — resolve the override before the argument is built. `data_dir_str` feeds all four sidecar spawn sites, so one change covers the plain, ROCm, CUDA and fallback paths.

Unset, empty and whitespace-only all mean "no override", so a variable cleared to `""` falls back to the platform path rather than the process working directory. Surrounding whitespace is trimmed, since variables set through the Windows GUI often carry a trailing space.

The startup log now names which source won, and the `create_dir_all` failure names the path, so a typo surfaces as an actionable error rather than a silent fallback.

## Behaviour note

The Tauri `data_dir` also anchors `backends/rocm`, `backends/cuda` and the `backend_override` file, so those move under the overridden directory too. That seemed right — the point of the variable is to put app data somewhere else — but flagging it in case you'd rather scope the override more narrowly.

## Tests

- `backend/tests/test_data_dir_env.py` covers the precedence rules on the Python side.
- Four unit tests in `main.rs` cover the override decision: unset, empty, whitespace-only, verbatim path, and trimming. `cargo test --bin voicebox` passes.

No UI changes, so no screenshots apply.

Fixes #981
Fixes #533

## Also partially addresses #577

#577 asks for the `Generated` and `Captured` folder locations to be configurable. Both sit under the data directory, so both move with this change — two of that issue's three asks. What it does *not* cover is downloading models straight to a chosen location (the issue's first point; `VOICEBOX_MODELS_DIR` plus the existing migration flow is the current answer), or a folder picker in Settings. Leaving #577 open on that basis rather than claiming it.

Refs #577

## Summary by CodeRabbit

* **New Features**
  * Configure the application’s data directory with the `VOICEBOX_DATA_DIR` environment variable.
  * Relative paths are resolved consistently, and surrounding whitespace is removed.
  * Explicit data-directory settings continue to take precedence over the environment variable.

* **Bug Fixes**
  * Improved fallback behavior when the variable is missing, empty, or whitespace-only.

* **Tests**
  * Added coverage for configuration, path resolution, defaults, trimming, and overrides.

### DIFF
--- a/backend/config.py
+++ b/backend/config.py
@@ -18,8 +18,10 @@ if _custom_models_dir:
     os.environ["HF_HUB_CACHE"] = _custom_models_dir
     logger.info("Model download path set to: %s", _custom_models_dir)
 
-# Default data directory (used in development)
-_data_dir = Path("data").resolve()
+# Default data directory (used in development). VOICEBOX_DATA_DIR lets a bare
+# `uvicorn backend.main:app` point at the packaged app's data dir without a CLI
+# flag. The --data-dir argument still wins: it calls set_data_dir() after import.
+_data_dir = Path(os.environ.get("VOICEBOX_DATA_DIR") or "data").resolve()
 
 
 def _path_relative_to_any_data_dir(path: Path) -> Path | None:
--- a/backend/tests/test_data_dir_env.py
+++ b/backend/tests/test_data_dir_env.py
@@ -15,14 +15,11 @@ reloads.
 """
 
 import importlib
-import sys
 from pathlib import Path
 
 import pytest
 
-sys.path.insert(0, str(Path(__file__).parent.parent))
-
-import config  # noqa: E402
+from backend import config
 
 
 @pytest.fixture
@@ -37,6 +34,10 @@ def reload_config(monkeypatch):
         return importlib.reload(config)
 
     yield _reload
+    # Undo before the reload, not after: pytest restores monkeypatch once the
+    # fixture has finished, so reloading first would bake the last test's
+    # VOICEBOX_DATA_DIR into the module for everything that follows.
+    monkeypatch.undo()
     # Leave the module matching the real process environment for later tests.
     importlib.reload(config)
 
--- a/backend/tests/test_data_dir_env.py
+++ b/backend/tests/test_data_dir_env.py
@@ -1,0 +1,78 @@
+"""
+Unit tests for the ``VOICEBOX_DATA_DIR`` environment variable.
+
+``backend/README.md`` documents ``VOICEBOX_DATA_DIR`` alongside ``--data-dir``,
+but the default was previously hardcoded to ``./data``, so a bare
+``uvicorn backend.main:app`` (how ``just dev`` starts the backend) always wrote
+to the repo instead of the app data dir.
+
+The variable is read at import time, so each test reloads the module under a
+patched environment.
+
+NOTE: These tests reload ``config``, which rebinds its module-global
+``_data_dir``. Import the module fresh rather than holding a reference across
+reloads.
+"""
+
+import importlib
+import sys
+from pathlib import Path
+
+import pytest
+
+sys.path.insert(0, str(Path(__file__).parent.parent))
+
+import config  # noqa: E402
+
+
+@pytest.fixture
+def reload_config(monkeypatch):
+    """Reload ``config`` with a patched environment, restoring it afterwards."""
+
+    def _reload(data_dir: str | None):
+        if data_dir is None:
+            monkeypatch.delenv("VOICEBOX_DATA_DIR", raising=False)
+        else:
+            monkeypatch.setenv("VOICEBOX_DATA_DIR", data_dir)
+        return importlib.reload(config)
+
+    yield _reload
+    # Leave the module matching the real process environment for later tests.
+    importlib.reload(config)
+
+
+def test_env_var_sets_data_dir(reload_config, tmp_path):
+    target = tmp_path / "appdata"
+    cfg = reload_config(str(target))
+
+    assert cfg.get_data_dir() == target.resolve()
+
+
+def test_defaults_to_local_data_dir_when_unset(reload_config):
+    cfg = reload_config(None)
+
+    assert cfg.get_data_dir() == Path("data").resolve()
+
+
+def test_empty_env_var_falls_back_to_default(reload_config):
+    cfg = reload_config("")
+
+    assert cfg.get_data_dir() == Path("data").resolve()
+
+
+def test_relative_env_var_is_resolved_to_absolute(reload_config):
+    cfg = reload_config("relative/data")
+
+    assert cfg.get_data_dir().is_absolute()
+    assert cfg.get_data_dir() == Path("relative/data").resolve()
+
+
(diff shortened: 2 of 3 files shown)


Lvigentini:
Verified end to end on Windows 11 against a build from this branch.

Set `VOICEBOX_DATA_DIR` to an empty `C:\Users\psyco\test_folder` and launched the built app. Startup log:

```
=================================================================
Starting voicebox-server sidecar
Data directory: "C:\Users\psyco\test_folder" (from VOICEBOX_DATA_DIR)
Remote mode: false
No ROCm backend found
No CUDA backend found, using bundled CPU binary
Sidecar command created successfully
Spawning bundled CPU server process...
```

`GET /health/filesystem` — the exact check from #981, which previously returned AppData paths:

```json
{"healthy": true, "directories": [
  {"path": "C:\Users\psyco\test_folder\generations", "exists": true, "writable": true},
  {"path": "C:\Users\psyco\test_folder\captures",    "exists": true, "writable": true},
  {"path": "C:\Users\psyco\test_folder\profiles",    "exists": true, "writable": true},
  {"path": "C:\Users\psyco\test_folder",              "exists": true, "writable": true}]}
```

The directory was populated with `backends/`, `captures/`, `generations/`, `profiles/` and a fresh 139 KB `voicebox.db`, and `GET /profiles` returned `0` — the app came up on an empty database rather than the existing library, which confirms it is genuinely reading from the override and not just logging it. The real data directory was untouched (`voicebox.db` still 1,839,104 bytes).

Also worth flagging from the same build: `just build` fails at the very end with

```
A public key has been found, but no private key.
Make sure to set `TAURI_SIGNING_PRIVATE_KEY` environment variable.
error: recipe `build-tauri` failed
```

Both installers are produced before that point, so it only affects the updater artifacts — unrelated to this PR, but it means a local `just build` exits non-zero without that secret.

Issue #1006, opened 2026-08-08
Stories: multi-track timeline editor with per-lane mixing and audio export


Lvigentini:
Opening this per CONTRIBUTING ("discuss major features in an issue first") to describe work that is ready and to check the design is wanted before review.

## Proposal

Turn the Stories tab into a multi-track timeline editor, so generated clips and imported audio can be assembled into a finished piece rather than played back one at a time.

**Tracks and mixing.** A `story_tracks` table holding per-lane name, volume, mute, solo and an optional duck-under-lane. Keyed by `(story_id, index)` rather than owning clips, so `story_items.track` stays a plain integer and drag/move/reorder are untouched — a lane with no row mixes at unity gain.

**Stereo, project-rate mixdown.** Today the mixer flattens every source to 24 kHz mono and sums into a 1-D buffer, so an imported music bed comes out at 12 kHz Nyquist with its stereo image folded flat. Instead: decode each source at its native rate, derive a project rate from the sources (capped at 48 kHz), and mix into a stereo buffer. Each lane renders separately, which is what makes ducking possible — a bed attenuated by the *finished* speech lane.

**Timeline editing.** Ripple move (off by default, since it rewrites clips the user did not touch), right-click for exact clip timing, per-clip fades and speed.

**Export.** WAV, MP3, OGG, Opus and FLAC via the bundled libsndfile — no ffmpeg needed. When ffmpeg *is* on PATH it adds EBU R128 loudness normalisation; without it the request still succeeds using peak normalisation.

**Folders for stories**, matching the clip/voice folders in #821.

## Possible relation to #715

#715 reports story audio being cut off during playback while the playhead is still in the waveform, and the cut-off audio blending with the next clip on export.

This work changes clip duration to come from the decoded array rather than `generation.duration`. A clip pinned to a version of a different length was mistimed, and a NULL duration raised `TypeError` inside a bare `except: continue` — silently dropping the clip from the export. That looks like the same failure, but I have not reproduced #715 directly, so I would rather flag it than claim a fix.

## Status

Implemented and pushed; PR to follow. Happy to split it (backend mixing / tracks schema / frontend editor) if that reviews better.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #870, opened 2026-07-09
REQUEST: Allow Editing of  segments in Stories


strmediallc:
Not sure if this is the proper place to put this, but I wanted to know if there was a way to edit the individual segments when creating a story? I noticed I had a few misspellings in various segments and wanted to know if there is already a way (or can there be a way) to edit an existing segment before regeneration?

While I'm still exploring, this is an amazing app. Looking forward to being a long-time user and supporter.

Continued success!

Lvigentini:
Flagging what the timeline work in #1007 does and doesn't cover here, since it's the PR closest to this.

It gives a segment trim, fades, speed, split, duplicate and version-swap — but **not** editing the text and regenerating in place, which is what you're asking for. I've listed it as explicitly out of scope in that PR rather than let it read as an oversight.

The reason it isn't a small addition: a story item points at a `generation_id`, so changing the text means generating a new clip and reattaching the item to it, while keeping the timing, fades and track placement you'd already set. That's a regenerate-and-reattach path, which is a different change from the mixing work.

Version-swap gets you part way today if the fix you need already exists as another take of that segment. For a typo it doesn't help.

Issue #715, opened 2026-05-29
Stories Audio doesn't match the waveform, the playback and the export


Argobelz:
When you arrange the audio and then play it, sometimes the audio is cut off while you see the playhead is still in the audio waveform. Then you try to export it, turns out the cutoff audio is blended together with the next audio in the timeline, and they speak together now.

Devin-Mattox:
I am experiencing this issue as well.

Issue #55, opened 2026-02-05
Feature request: add pauses via code in Story


akademiebios:
Can you add a feature in the story mode to add pauses via code snippets as we can do it in Elevenlabs?
E.g. <break time="1.0s" /> or ... or [pause]
This way we could adjust our texts to have breaks wherever we need them (and it should be easy to implement). 

Maybe also an option to set the speed of the spoken text. Something like <speed="0.75" /> (for 75% of the normal speed = slower)

DonRichards:
I support this request. If you're considering adding inline narration controls, it might be worth expanding the scope beyond speed changes to support more general performance directions.

For example, instead of only supporting something like `[speed]`, consider a small set of expressive cues such as:

- **Pauses**
  - `[pause]`
  - `[pause=2s]`

- **Speaking rate**
  - `[rate=0.75]`

- **Vocal delivery**
  - `[whisper]`
  - `[softly]`
  - `[emphasis]`

- **Non-verbal cues**
  - `[sigh]`
  - `[clears_throat]`
  - `[laughs]`
  - `[gasps]`

- **Voice quality**
  - `[trembling]`
  - `[breathless]`

These map well to concepts already used in speech synthesis (e.g. SSML's `<break>` and `<prosody>`), so there are existing conventions to draw from without necessarily exposing SSML directly to users.

Even if the initial implementation only supports pauses and speaking rate, designing the syntax with future expressive narration in mind could make it easier to extend later without introducing a second markup format.

## Add note:
It may also make future compatibility with SSML or other TTS backends easier if the syntax is designed with those concepts in mind.

DonRichards:
## Related issues:
- https://github.com/jamiepine/voicebox/issues/739
- https://github.com/jamiepine/voicebox/issues/771

Lvigentini:
Same note as on #870: the timeline work in #1007 doesn't cover this, and I've marked it out of scope there rather than leave it ambiguous.

That PR works at the mixer level — it can place silence between clips by moving them apart on the timeline, and right-click gives exact start/end times, so a pause *between* segments is achievable by hand. What it can't do is a pause *inside* one, which is what inline markup is for. That belongs at the text layer, before generation, not in the mix.

[...]

PR #1007, opened 2026-08-08
feat(stories): multi-track timeline editor, per-lane mixing, and audio export

### LINKED ISSUE #869: Any chance to get .mp3 from the API?
It wouild be great if it /audio supported something like a format-parameter where one could pass the value "mp3" to get a that directly.

Lvigentini:
Implements #1006.

## What it does

- **Tracks** — new `story_tracks` table with per-lane name, volume, mute, solo and optional duck-under-lane. Keyed by `(story_id, index)`, so `story_items.track` stays a plain integer and drag/move/reorder are untouched; a lane with no row mixes at unity gain.
- **Stereo, project-rate mixdown** — replaces the 24 kHz mono flatten. Sources are decoded at native rate, the project rate is derived from them (capped at 48 kHz), and each lane renders to its own buffer, which is what makes ducking possible.
- **Timeline editing** — ripple move (off by default), right-click for exact clip timing, per-clip fades and speed.
- **Export** — WAV, MP3, OGG, Opus, FLAC through the bundled libsndfile; no ffmpeg required. ffmpeg, when present, adds EBU R128 loudness normalisation.
- **Story folders**, matching the clip/voice folders.

Backend: `routes/stories.py`, `services/stories.py`, `services/history.py`, plus `backend/tests/test_story_mixdown.py`. Frontend: `StoryTrackEditor`, `TrackMixerControls`, `StoryContent`, `StoryList`, `StoryChatItem`, `useStories`.

## Also fixes #869

#869 asks for a `format` parameter on `/audio` so API callers can get MP3 without transcoding themselves. The encoder that needs — `EXPORT_FORMATS` and `encode_audio` in `utils/audio.py` — is already in this branch for story export, so the endpoints only had to reach for it.

`GET /audio/{generation_id}` and `GET /audio/version/{version_id}` now take `?format=wav|mp3|ogg|opus|flac`. Omitted, the stored file is streamed by `FileResponse` exactly as before, so existing callers and range requests are untouched; the transcode buffers a whole encode, so it is only paid for when asked for. A file already in the requested container is handed back as-is rather than round-tripped. An unsupported container is a 400 naming what is on offer, not a silent fallback to WAV.

`backend/tests/test_audio_format_param.py` — 10 tests covering the default path, the same-container shortcut, each container, sample-rate preservation, the Opus 48 kHz resample, and both rejections.

Fixes #869

## Possible fix for #715

#715 reports clips cut off in playback and blending on export. This changes clip duration to come from the decoded array rather than `generation.duration` — a clip pinned to a version of a different length was mistimed, and a NULL duration raised `TypeError` inside a bare `except: continue`, silently dropping the clip from the export. Same shape as the report, but I have not reproduced #715 directly, so flagging rather than claiming. Refs #715.

## Stacking

Carries the folder work from #1005, including its review-fix commit (`b74eb6a`, cherry-picked here — this branch had drifted behind #1005 and was showing already-fixed bugs). It also carries `d5eea00`, the backend half of #1004, but **not** #1004's two Tauri commits.

So the branches aren't strict ancestors of one another — they're independent branches that share commits. Verified with `git cherry`: #1007 now contains every patch in #1005. Merge order #1004 → #1005 → this still reads cleanest, but nothing breaks in another order.

## Notes for review

- Screenshots to follow from the author — this is a substantial UI change.
- 48 files, but ~13 focused commits; reviewing commit-by-commit will read better than the squashed diff.
- Deliberately excluded from this branch: a version bump and hand-edits to `CHANGELOG.md`, since releases and that file are yours to manage.

## Out of scope, on purpose

**#870 — editing a segment's text.** This PR gives a segment trim, fades, speed, split, duplicate and version-swap, but no way to fix a typo and regenerate in place. That's the most obvious remaining hole in the timeline, and I'd rather flag it than have it read as an oversight: it needs a regenerate-and-reattach path through the item's `generation_id`, which is a different change from the mixing work here. Happy to take it as a follow-up if you want it.

**#55 — inline pause/prosody markup in stories.** Same reasoning; it belongs at the text layer, not the mixer.

## Screenshots

**Stories list and audio export**

Screenshot 2026-08-07 103615

The Stories tab with the story list, story content, and the `Export Audio` format picker: WAV, MP3, OGG Vorbis, Opus and FLAC all written by the bundled libsndfile with no ffmpeg required. The separate `Mastering` entry — `MP3, loudness normalised` — is the EBU R128 path, which is the one option that needs ffmpeg on PATH; without it the request still succeeds using peak normalisation.

**Multi-track timeline editor**

Screenshot 2026-08-09 005959

The track editor with several lanes and clips. Note the ruler alignment: `LABEL_COL_WIDTH` now drives the ruler and scrollbar corner spacers, fixing the 182 px offset described in f331d30.

**Per-lane mixer and ducking**

Screenshot 2026-08-09 012012

Per-lane volume, mute, solo and `Duck under track`, which attenuates a lane against the finished output of another — how a music bed sits under narration.

## Summary by CodeRabbit

* **New Features**
  * Organize voices, stories, and generation history into folders with nesting, drag-and-drop, filtering, search, renaming, and deletion.
  * Duplicate and preview voice profiles, with card or list display options.
  * Edit story fades, playback speed, track placement, and mixer controls.
  * Export audio as WAV, MP3, OGG, Opus, or FLAC with optional loudness normalization.
  * Clear searches with Escape or a dedicated button.
* **Bug Fixes**
  * Added clearer validation when required audio support is unavailable.
* **Documentation**
  * Updated API, audio export, configuration, and optional FFmpeg documentation.

### DIFF
--- a/app/src/components/History/ClipFolderTree.tsx
+++ b/app/src/components/History/ClipFolderTree.tsx
@@ -149,6 +149,13 @@ export function ClipFolderTree({
         {/* Shaded and bold, matching the voice folders, so a folder never
             reads as just another row in the list. */}
         <div
+          onDragEnter={(e) => {
+            // Both enter and over must preventDefault; some engines only treat
+            // an element as a drop target once enter has been cancelled.
+            if (!onDropItem || !isFolderDrag(e)) return;
+            e.preventDefault();
+            setDragOverId(folder.id);
+          }}
           onDragOver={(e) => {
             if (!onDropItem || !isFolderDrag(e)) return;
             // preventDefault is what marks this a valid drop target.
@@ -279,6 +286,11 @@ export function ClipFolderTree({
       <button
         type="button"
         onClick={() => onSelect({ kind: 'uncategorised' })}
+        onDragEnter={(e) => {
+          if (!onDropItem || !isFolderDrag(e)) return;
+          e.preventDefault();
+          setDragOverId(UNCATEGORISED_DROP_ID);
+        }}
         onDragOver={(e) => {
           if (!onDropItem || !isFolderDrag(e)) return;
           e.preventDefault();
--- a/app/src/components/History/ClipFolderTree.tsx
+++ b/app/src/components/History/ClipFolderTree.tsx
@@ -124,12 +124,14 @@ export function ClipFolderTree({ selection, onSelect }: ClipFolderTreeProps) {
 
     return (
       <div key={folder.id}>
+        {/* Shaded and bold, matching the voice folders, so a folder never
+            reads as just another row in the list. */}
         <div
           className={cn(
-            'group/node flex items-center gap-1 rounded pr-1',
-            isSelected && 'bg-accent/40',
+            'group/node my-0.5 flex items-center gap-1 rounded border border-border/60 bg-muted/60 pr-1',
+            isSelected && 'border-accent/60 bg-accent/40',
           )}
-          style={{ paddingLeft: `${depth * 12}px` }}
+          style={{ marginLeft: `${depth * 12}px` }}
         >
           {children.length > 0 ? (
             <button
@@ -151,8 +153,8 @@ export function ClipFolderTree({ selection, onSelect }: ClipFolderTreeProps) {
             onClick={() => onSelect({ kind: 'folder', folderId: folder.id })}
             className="flex min-w-0 flex-1 items-center gap-1.5 py-1 text-left"
           >
-            <span className="truncate text-xs">{folder.name}</span>
-            <span className="shrink-0 text-[10px] text-muted-foreground/70">
+            <span className="truncate text-xs font-bold text-foreground">{folder.name}</span>
+            <span className="shrink-0 rounded bg-background/70 px-1 text-[10px] font-semibold tabular-nums text-muted-foreground">
               {folder.item_count}
             </span>
           </button>
--- a/app/src/components/History/ClipFolderTree.tsx
+++ b/app/src/components/History/ClipFolderTree.tsx
@@ -281,7 +281,12 @@ export function ClipFolderTree({
         {allLabel ?? t('folders.clip.allClips')}
       </button>
 
-      {tree.map((node) => renderNode(node, 0))}
+      {/* Folders scroll on their own. With a few dozen folders the list
+          otherwise pushes the clips off-screen entirely, and "All clips" and
+          "Uncategorised" stay pinned outside it so both are always reachable. */}
+      <div className="max-h-[40vh] min-h-0 overflow-y-auto overscroll-contain">
+        {tree.map((node) => renderNode(node, 0))}
+      </div>
 
       <button
         type="button"
--- a/app/src/components/History/ClipFolderTree.tsx
+++ b/app/src/components/History/ClipFolderTree.tsx
@@ -1,0 +1,318 @@
+import {
+  ChevronDown,
+  ChevronRight,
+  FolderPlus,
+  Inbox,
+  Layers,
+  MoreHorizontal,
+  Pencil,
+  Trash2,
+} from 'lucide-react';
+import { useMemo, useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import { Button } from '@/components/ui/button';
+import {
+  Dialog,
+  DialogContent,
+  DialogDescription,
+  DialogFooter,
+  DialogHeader,
+  DialogTitle,
+} from '@/components/ui/dialog';
+import {
+  DropdownMenu,
+  DropdownMenuContent,
+  DropdownMenuItem,
+  DropdownMenuSeparator,
+  DropdownMenuTrigger,
+} from '@/components/ui/dropdown-menu';
+import { Input } from '@/components/ui/input';
+import type { FolderResponse } from '@/lib/api/types';
+import {
+  useCreateFolder,
+  useDeleteFolder,
+  useDetachFolder,
+  useFolders,
+  useUpdateFolder,
+} from '@/lib/hooks/useFolders';
+import { cn } from '@/lib/utils/cn';
+import { useUIStore } from '@/stores/uiStore';
+
+/**
+ * What the clip list is currently filtered to.
+ *
+ * `uncategorised` is its own selection rather than a null folderId, because
+ * "no filter" and "clips in no folder" are different requests — the server
+ * distinguishes them too.
+ */
+export type ClipFolderSelection =
+  | { kind: 'all' }
+  | { kind: 'uncategorised' }
+  | { kind: 'folder'; folderId: string };
+
+interface ClipFolderTreeProps {
+  selection: ClipFolderSelection;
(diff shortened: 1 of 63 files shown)


Lvigentini:
Thanks @coderabbitai — went through the review. Two of the findings were real and are fixed in f331d30; one I think is a false positive.

**Fixed**

- **`StoryTrackEditor.tsx` — 182 px ruler offset.** Confirmed: `LABEL_COL_WIDTH` is 246 while both corner spacers were still `w-16` (64 px). Both now derive from the constant so they can't drift again.
- **`useFolders.ts` — story-folder invalidation.** Confirmed: `FolderKind` has three values and the ternary only handled two, so deleting a story folder invalidated `['history']` instead of `['stories']`. Replaced with a `Record<FolderKind, string>` map, which makes a future fourth kind a type error rather than a stale list.

**Not applying: `folders_router` registration order**

The concern was that `PUT /profiles/{profile_id}` would shadow `PUT /profiles/{profile_id}/folder`. It can't — Starlette compiles that path to:

```
^/profiles/(?P<profile_id>[^/]+)$
```

`[^/]+` cannot cross a slash, so it matches `/profiles/abc` but not `/profiles/abc/folder`. Registration order is irrelevant here. `profiles.py` already declares `/profiles/{profile_id}/channels` and `/profiles/{profile_id}/effects` after the same parent route on `main` today, and both work.

The remaining comments are noted — the debounced volume slider in particular looks worth doing, and I'll follow up separately rather than grow this PR.

Lvigentini:
@coderabbitai — triaged all 18 findings. Two were live and are fixed in `3460a59`; most of the rest were already addressed by later commits on the branch, so they're reading as open when they aren't.

**Fixed now:**

- **`duck_under_track` unbounded** — real. A negative lane index was accepted and then silently never matched a lane at mix time, so the setting looked saved and did nothing. Bounded with `ge=0`. The *self*-ducking half of that finding was already handled in the route, which is where the lane's own index is known.
- **Exception chain** — `duplicate_profile`'s `ValueError` was re-raised without `from e`.

**Already fixed, in commits after the review:**

| Finding | Fixed by |
|---|---|
| Ruler/scrollbar `w-16`, 182px offset | `2b1b979` |
| Volume slider writes per step | `9a1a328` |
| Ducking depends on dict order | `9a1a328` |
| Timing dialog / duplicate / preview failures shown | `9a1a328` |
| Reject whitespace-only duplicate names | `b74eb6a` |
| `useDeleteFolder` wrong cache for story folders | `b74eb6a` |
| Collapse toggle action label | `4530217` |

`b74eb6a` is worth calling out: this branch had drifted behind #1005 and was missing its review-fix commit entirely, so several findings here were duplicates of ones already answered there. Cherry-picked in `2c7050e`; `git cherry` now confirms #1007 contains every patch in #1005.

**Assessed and not taken — `Register folders_router before profiles_router`.** I don't think this one is real. `folders.py` registers `PUT /profiles/{profile_id}/folder`, which is a distinct path from `profiles.py`'s `/profiles/{profile_id}` rather than a shadowed one; FastAPI matches full path patterns, and `test_folders.py` exercises the endpoint across 59 passing tests. Happy to be shown otherwise if there's a case I'm missing.

Still open and worth doing, just not in this commit: atomic profile-name allocation (a real check-then-act, same shape as the one I just fixed in #1025), the `soundfile` dependency pin, the drag-event ordering in `FolderSection`, and the README wording.

Lvigentini:
Follow-up note for whenever this lands: #1026 adds the backend half of #870 — `POST /generate/{id}/regenerate` now takes `text`/`language`/`instruct`/`seed` overrides, and `generation_versions` records what produced each take, so a corrected segment keeps its version history instead of needing a new generation.

What it deliberately leaves out is the part that belongs here: when a correction changes a segment's length, the following clips on that lane need to ripple by the delta, or the arrangement drifts. This PR already has ripple move, so it's the right home for it — I kept it out of #1026 rather than write timeline code against `main` that would conflict with this branch.

Once both are in, the flow is: edit text → regenerate → item repins to the new version → lane ripples by the duration delta, with keep-timing as the opt-out.

Lvigentini:
The three findings I'd left open are now done in `b0b865c`, so this review is fully worked through.

**Name allocation was check-then-act.** `get_unique_profile_name` only SELECTs and the INSERT lands later, so two concurrent duplicates or imports could both be handed the same name — the loser then hit the UNIQUE constraint and surfaced as a 500 instead of taking the next suffix. `insert_profile_with_unique_name` settles the name *by* inserting it and retries on `IntegrityError`, so the constraint arbitrates.

Two details worth flagging for review: `build_row` is a callback because a rolled-back commit expunges the instance, so each attempt needs a fresh one. And it runs before any file work — retrying after the sample copies would mean undoing them, whereas the profile directory is keyed by the new id rather than the name, so a name retry doesn't touch it. `get_unique_profile_name` stays for callers that need a candidate name before they have a row, with its raciness now documented rather than implied.

**The drop target ran the browser default on rejected payloads.** Correct and the more serious of the two UI ones — `onDragOver` had already accepted the element, so returning early from `onDrop` let a dropped file or URL navigate the webview away. Cancel first, decide second. `onDragLeave` also fired on every crossing into the header's own children, so the highlight flickered for the whole drag; only a leave of the header itself counts now.

**The `soundfile` floor didn't match the promise.** `EXPORT_FORMATS` documents MP3 and Opus with no ffmpeg, which holds only from the wheel bundling libsndfile 1.2.2 — the floor was `>=0.12.0`, whose bundled 1.1.0 can write neither. Raised to `>=0.13.0` in both requirements files with the reason recorded next to it. Verified locally: 0.14.0 / libsndfile 1.2.2, MP3 write and Opus both available.

191 backend tests pass, frontend typecheck clean, no new lint.

Lvigentini:
Pushed `7bc2ca7`, a fix to per-clip speed that came out of listening tests on other work.

**Speed used a phase vocoder.** `librosa.effects.time_stretch` reconstructs from magnitudes and re-estimated phase. On speech that smears consonants and leaves a phasey ring — distinct enough that a listener rejected it outright when I was testing the same algorithm elsewhere. Pitch survives either way; only the artefacts differ.

Replaced with WSOLA in `utils/audio.py`: stays in the time domain, overlap-adds real waveform segments, picks each splice by cross-correlation so pitch periods line up. Nothing is resynthesised. numpy only, no new dependency.

**Stereo was broken in a way the loop hid.** The obvious implementation — stretch each channel — is wrong twice, because the splice search is content-dependent:

- left and right choose *different* splice points, so the stereo image tears;
- the channels come out different lengths, and `np.stack` rejects that outright.

The mixer works in `(channels, samples)` throughout, so this is the shape it actually hits. Every story clip with a speed adjustment was going through it.

Splices are now planned once from the mono downmix and applied to every channel, so all channels cut identically, the image survives, and lengths match by construction. Found by a test asserting stereo keeps its shape — the naive version failed it immediately.

6 tests: ratio accuracy across four rates, pitch preservation via dominant-frequency comparison before and after, and stereo keeping shape and length.

197 backend tests pass, no new lint.

PR #1025, opened 2026-08-09
feat(tts): add a pronunciation dictionary

### LINKED ISSUE #827: Feature Request: Add pronunciation dictionary support
Voice cloning and TTS quality are excellent, but there is currently no reusable way to correct pronunciations for names, acronyms, brands, or technical terms. Users must manually edit the text each time or rely on engine-specific phoneme syntax.

Lvigentini:
Closes #827.

Names, acronyms, brands and loanwords come out wrong and there is no reusable way to fix them — you edit the text every time, in every script. @aferreira-deo makes the same point on that issue about WCAG and technical terms in coding tutorials.

## Approach

A `term -> respelling` map applied just before TTS:

```bash
curl -X POST localhost:17493/pronunciations \
  -d '{"term": "bandeja", "replacement": "ban-DEH-ha", "language": "en"}'
```

**Respelling rather than phonemes, on purpose.** Every engine reads plain text, so this works on all eight; a phoneme string only works on the engines that accept one. Cruder, portable. Phoneme support can layer on later per engine without changing the model.

**Scoping.** Entries are global by default, and can be narrowed to a language, a voice, or both. Language scope is the one that earns its keep for mixed-language work: a Spanish term needs respelling while the engine is reading English, and must be left alone when it is already reading Spanish. A profile-scoped entry beats a global one; a language-specific entry beats a wildcard.

## The stored text stays clean

The dictionary runs at generation time, not when the text is saved. `generations.text` keeps what the author wrote, so History never shows a reader `ban-DEH-ha`, and editing an entry changes future audio without rewriting the past.

That is also why `POST /pronunciations/preview` exists — the rewritten string is never stored, so without it there is no way to see what a rule actually does short of listening to the output.

## Matching

A single pass over one alternation of every term, longest first. That is what stops replacements cascading: with `bandeja -> ban-DEH-ha` and `ha -> hah`, a loop of per-term substitutions yields `ban-DEH-hah`. Longest-first also lets `bandeja alta` beat the `bandeja` inside it.

- Word boundaries via lookarounds rather than `\b`, so terms with punctuation still anchor
- Terms are escaped — a term is text, not a pattern
- `[laugh]`-style tags are skipped; they are engine syntax, not speech
- Capitalisation carries onto the replacement, counting *cased* characters rather than `str.isupper()` — that returns True for `C++`, and shouting the replacement would give `C PLUS PLUS`

Duplicate scopes are rejected in the service rather than by a unique constraint, since SQL treats NULLs as distinct and a constraint would happily accept two global entries for the same term.

Applied on both `/generate` and `/generate/stream`, so a streamed preview matches what the persisted path produces.

## Endpoints

| Method | Path | |
|---|---|---|
| GET | `/pronunciations` | list, optionally filtered to what applies to a language/voice |
| POST | `/pronunciations` | create (409 on a duplicate scope) |
| PUT | `/pronunciations/{id}` | update; omitted fields are left alone |
| DELETE | `/pronunciations/{id}` | delete |
| POST | `/pronunciations/preview` | see the rewrite without generating |

## Tests

28 in `backend/tests/test_pronunciation.py` — matching, the no-cascade rule, capitalisation including the `C++` case, scope resolution and leakage, degenerate input, CRUD, and the property the design rests on: the engine receives the respelling and the stored row does not. The model is mocked there; loading 3.5 GB of weights would not make the assertion any truer.

Backend only — no UI yet, deliberately, so the model and matching rules can be argued about before anything is built on them.

Branched off `main`, independent of my other open PRs.

**Unrelated observation while testing:** on `main` the backend suite writes into the repo's `./data` because `VOICEBOX_DATA_DIR` is not honoured — every test module that sets it (`test_story_mixdown.py`, `test_data_dir_env.py`, and others) is silently using the real data directory, so state persists between runs. #1004 fixes that; flagging since it makes the suite order-dependent today.

## Summary by CodeRabbit

* **New Features**
  * Added a pronunciation dictionary for custom term respellings.
  * Added tools to create, update, delete, filter, enable, disable, and preview pronunciation entries.
  * Added language- and profile-specific pronunciation settings.
  * Applied pronunciation substitutions during speech generation while preserving the original text in history.
  * Preserved capitalization, avoided chained replacements, and prevented duplicate pronunciation entries.
* **Tests**
  * Added comprehensive coverage for pronunciation matching, validation, scoping, previews, and generation behavior.


Lvigentini:
Thanks @coderabbitai — four of five were real. Addressed in `425dbc8`.

**`ilike` treated terms as patterns.** The worst of them. `find_duplicate` used `term.ilike(term)`, so `%` and `_` were wildcards rather than literals: `band_ja` collided with `bandeja` and was rejected as a duplicate. Now compares `lower(term)`, trimmed first — the route stores the trimmed value, so an untrimmed lookup was missing its own duplicate too.

**Whitespace-only values.** Correct, and there was already a `TrimmedName` precedent in the folders work I should have reused. Both request models now strip before length-checking.

**Preview accepted an unknown profile.** Correct — it fell back to global scope and reported a result generation wouldn't produce. Validates first now.

**Scope uniqueness.** You were right and my code comment was wrong. It claimed a constraint couldn't express this because SQL treats NULLs as distinct — true of a plain `UniqueConstraint`, but a unique *expression* index handles it:

```sql
CREATE UNIQUE INDEX uq_pronunciation_scope ON pronunciation_entries (
    lower(term), COALESCE(language, ''), COALESCE(profile_id, '')
)
```

`COALESCE` maps the wildcard scopes onto comparable values, so two global entries for the same term now collide as they should. Violations map to 409 instead of 500; `find_duplicate` stays as the early check that can name the existing row.

That needed a migration you didn't ask for but which the change implies: `create_all` builds the index with the table, but won't add one to a table that already exists, so a database from an earlier build of this branch would never get it. Pre-existing duplicates are collapsed first (oldest row per scope) or `CREATE UNIQUE INDEX` fails on them.

Worth recording: the first version used plain `CREATE UNIQUE INDEX` guarded by an inspector check, and the full suite caught it — several test modules boot the app in one process, the inspector reflects a stale snapshot, and the migration raised. A migration that can raise takes startup down with it, so it's `IF NOT EXISTS` now with the inspector check kept only as the fast path that skips the dedup scan.

**Not taken — logging terms at INFO.** Fair point that terms can be names, so the values moved to DEBUG and INFO carries only the count. I didn't go further and drop them entirely: when a rewrite produces wrong audio, "which rule fired" is the first question, and DEBUG is the right level for that rather than nowhere.

7 further tests covering each, including the case-differing global pair a plain `UNIQUE` would have let through.

Lvigentini:
Both correct, fixed in `4c2ad09`.

**`MIN(id)` is not "oldest".** Good catch — the code didn't do what its own comment claimed. The dedup step said it kept the oldest row per scope and then ordered by `MIN(id)` over random UUIDs, so it kept an arbitrary row, and a *different* arbitrary row on a different machine.

Now ordered by `COALESCE(created_at, '') || '|' || id` — timestamp first, id only to break ties deterministically. ISO-8601 text sorts chronologically, so this avoids a window function and stays portable across SQLite builds rather than assuming 3.25+.

Added a test that inserts the newer row *first*, so insertion order can't be what makes it pass.

**The privacy assertion was weak.** Right — it checked `"Alicia"` only, so a message leaking just the surname would have passed a test whose whole point is that names don't reach INFO. Now asserts on every fragment of the term and on the replacement.

PR #1026, opened 2026-08-09
feat(api): regenerate a take with corrected text or settings

Lvigentini:
Refs #870.

That issue asks to fix a typo in a story segment and re-run it. Today that means building a whole new generation — which loses the timeline placement, fades, speed and track the old one had. A one-character correction costs the arrangement around it.

## Most of this already existed

`POST /generate/{id}/regenerate` already saved its output as a new **version** under the same generation id, and story items can already be pinned to a version. It just couldn't be told to change anything — it re-ran `gen.text` verbatim.

It now takes an optional body:

```bash
curl -X POST localhost:17493/generate/{id}/regenerate \
  -d '{"text": "He plays a bandeja, not a smash."}'
```

`text`, `language`, `instruct`, `seed` — omitted fields reuse the generation's settings, and **no body at all behaves exactly as before**.

## Versions become self-describing

`generation_versions` held only `audio_path` and `effects_chain`, so a corrected take was unattributable: the generation row would describe the newest text while older takes still pointed at audio of the *old* text, with nothing recording which was which.

Four nullable columns fix that. **NULL means "same as the generation"** — so every existing row is already correct and needs no backfill, and only a field the caller actually overrode is written. That keeps "unchanged" and "explicitly set to the same value" distinguishable rather than collapsing them.

**The generation row is never rewritten.** It keeps the text as first written; each take records what produced it. The take with the typo stays playable *and* attributable instead of becoming audio nobody can account for — which is what makes A/B between takes mean anything.

## Seed

A regenerate normally drops the seed so the take varies. An explicitly passed seed now survives, since that's the caller asking for a specific result — usually to reproduce a take they liked.

## Deliberately not included

The timeline half of #870 — rippling the following clips by the duration delta when a correction changes a segment's length — isn't here. It belongs in the story routes, which #1007 is currently rewriting; adding it on top of `main` would conflict for no benefit. It's the natural follow-up once #1007 lands, which is also why this says `Refs` rather than `Closes`.

## Tests

12 in `backend/tests/test_regenerate_overrides.py`: corrected text reaching the engine, the generation row keeping the original, the take recording its own settings, unoverridden fields staying NULL, plain regenerates recording nothing, takes accumulating rather than replacing, both seed rules, and the API surface including no-body compatibility.

The model is mocked — what's under test is which strings go where, and loading 3.5 GB of weights wouldn't make those assertions any truer.

Branched off `main`, independent of my other open PRs.

## Summary by CodeRabbit

* **New Features**
  * Regenerate audio with optional text, language, instruction, and seed overrides.
  * Preserve and display generation-specific settings for each audio version.
  * Support explicit seed preservation during regeneration.

* **Bug Fixes**
  * Improved validation for invalid languages, empty text, and missing generations.
  * Maintained compatibility with regenerations that provide no override settings.

### DIFF
--- a/backend/database/migrations.py
+++ b/backend/database/migrations.py
@@ -201,6 +201,17 @@ def _migrate_generation_versions(engine, inspector, tables: set[str]) -> None:
     if "source_version_id" not in columns:
         _add_column(engine, "generation_versions", "source_version_id VARCHAR", "source_version_id")
 
+    # Nullable with no default: NULL means "same as the generation", so
+    # existing takes are already correct and need no backfill.
+    for column, ddl in (
+        ("text", "text TEXT"),
+        ("language", "language VARCHAR"),
+        ("instruct", "instruct TEXT"),
+        ("seed", "seed INTEGER"),
+    ):
+        if column not in columns:
+            _add_column(engine, "generation_versions", ddl, column)
+
 
 def _migrate_capture_settings(engine, inspector, tables: set[str]) -> None:
     if "capture_settings" not in tables:
--- a/backend/database/models.py
+++ b/backend/database/models.py
@@ -138,6 +138,20 @@ class GenerationVersion(Base):
     effects_chain = Column(Text, nullable=True)
     source_version_id = Column(String, ForeignKey("generation_versions.id"), nullable=True)
     is_default = Column(Boolean, default=False)
+
+    # What produced this take, when it differs from the generation's own
+    # settings. NULL means "same as the generation" -- so every existing row
+    # stays correct without a backfill, and only an overridden regenerate
+    # writes anything here.
+    #
+    # Without these a corrected take is unattributable: the generation row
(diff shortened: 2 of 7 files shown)


Lvigentini:
Correct, and it was the worst shape a test can have. Fixed in `b46020e`.

`assert r.status_code in (200, 400)` passed when the contract it exists to protect was broken — the generation is created `completed`, so a 400 there means the no-body path failed. The hedge made the assertion unfalsifiable for its own subject.

Tightening it to `== 200` then exposed a second problem you didn't ask about but which the change surfaced: the route enqueues the work and returns, so the background task outlives the test and its monkeypatched backend, and then reaches for the real 3.5 GB model. The file went from 5s to 39s. Cancelling the generation after asserting the response brings it back to 6s — the response is what this test covers, and the generation path itself is asserted in the tests above it.

PR #1036, opened 2026-08-10
feat(prosody): dictionary strategies, LLM annotation, and generation wiring

Lvigentini:
Builds on #1035 (the prosody library) and #1025 (the pronunciation dictionary). This is the half that changes behaviour, kept separate so #1035 can be judged without it.

**Stacking:** this branch carries #1025's three commits and sits on #1035's six, so it shows both until they merge. Merge order #1035 → #1025 → this, though nothing breaks in another order.

## The dictionary becomes a source of markup

An entry no longer has a private substitution pass. It emits the same directive an author would have typed:

```
bandeja  →  <sub alias="bandeha">bandeja</sub>
víbora   →  <lang xml:lang="es">víbora</lang>
```

and the ordinary parse–compile–render pipeline takes it from there. That buys three things at once: dictionary terms and hand-written spans **compose**, because by the time the compiler sees them they are the same thing; every rule the compiler enforces applies to dictionary output for free; and the result is **showable**, so a rule firing unexpectedly is visible rather than an invisible gap between what was typed and what was spoken.

Three strategies, with `replacement` always populated as the fallback when the preferred one is unavailable on the target engine:

| strategy | markup | works on |
|---|---|---|
| `respell` | `<sub alias="bandeha">` | every engine |
| `language` | `<lang xml:lang="es">` | every engine, via cutting |
| `phoneme` | `<phoneme ph="…">` | engines that accept it |

A strategy that cannot be realised is rejected at the API rather than stored — `strategy="language"` with no `spoken_language` would show the user the strategy they picked while silently falling back.

## LLM annotation, never in the render path

The local LLM reads a script and returns the same script with directives inserted. From there the deterministic pipeline runs exactly as for hand-typed markup — which is what keeps "with an LLM" and "without an LLM" the same code path downstream, and why generation stays reproducible. A model in the render path would make the same script produce different audio every run.

**The invariant that makes it safe:** strip the tags from the model's answer and compare to the input. If a single word moved, the model rewrote the script instead of annotating it and the answer is discarded. It can fail to help; it cannot mangle.

Small models wrap answers in fences and quotes. Rather than guess which, every plausible unwrapping is offered and the invariant picks the first faithful one — so no unwrapping can launder a changed word into acceptance.

**Without an LLM it degrades, not breaks.** A missing model is reported, never downloaded on demand; `GET /prosody/annotate/availability` lets a client hide the action rather than offer something that will fail. Annotation is typing assistance — everything else works without it.

## Wiring

`POST /generate` and `/generate/stream` now resolve dictionary entries and markup before synthesis, through one shared function so "unmarked text behaves as before" is a property of a single place rather than a claim repeated twice.

The property under most scrutiny is the one about *not* changing anything:

- plain text takes the same single-shot call, same arguments;
- `5 < 6` stays literal — the parser only knows a closed tag set;
- malformed markup falls back to speaking the text literally rather than failing the generation.

Auto-detected rather than opt-in, with `prosody: false` as the escape hatch for a script that genuinely contains something tag-shaped. An opt-in flag would have meant updating every caller — MCP, API, story regeneration — or leaving the feature invisible.

The stored row keeps the **markup**, not the resolved text, consistent with #1025 and #1026: the resolved form is derivable, the author's is not, and editing markup to regenerate needs it intact.

## `POST /prosody/preview`

Compiles a script to its plan with no audio: every cut, language and silence, which dictionary terms fired, the markup they resolved to, and everything the chosen engine cannot honour. Free, because the plan is plain data.

## Two bugs found while building it

`is_trivial` excluded any plan carrying a substitution, so every respelled sentence would have taken the renderer path for nothing — contradicting the property that makes respelling preferred, that it does not cut.

The migration guard for `uq_pronunciation_scope` never fired: SQLAlchemy cannot reflect an expression-based index, so the inspector never reported it and the dedup scan ran on every startup. Reads `sqlite_master` directly now.

## Tests

209 on top of #1035's 93 — annotator, LLM invariant with the model stubbed, preview, and the pipeline integration whose largest section asserts that unmarked text is untouched.

No UI yet, deliberately: dictionary and prosody are API-only so the model and the matching rules can be argued about before anything is built on them.

## Summary by CodeRabbit

* **New Features**
  * Added pronunciation dictionary management with scoped entries, replacement strategies, validation, and preview support.
  * Added prosody markup previews for pauses, languages, speaking rates, emphasis, substitutions, and phonemes.
  * Added optional local AI-assisted prosody annotation with availability checks and validation.
  * Added prosody-aware speech generation with engine capability handling and improved audio transitions.
* **Enhancements**
  * Generation now supports enabling or disabling pronunciation and prosody processing.
  * Added warnings for unsupported language or speech features during previews.


hakimio:
@Lvigentini

Proposed simple UI. Code [here](https://github.com/hakimio/voicebox/tree/feat/prosody-user-facing)

image

image

image

image

Lvigentini:
@hakimio — Tomas, this is the most useful thing that's happened to this PR. Thank you for taking the time to build on an unmerged branch; that's a real cost to take on, and you didn't just add a UI on top, you found four genuine defects in the code underneath it. I've adopted all four and built your UI out against the rest of the stack.

Below is the full account: what I took, why it convinced me, and — since I changed things on the way in — exactly what I altered, so nothing is a surprise when you read the diff.

## The parser fix is the best find in the branch

You framed it as an innocent word raising a parse error. That's true and would have been enough, but while writing the regression test I found it's worse:

```
'Fill in the <subject> field.'   → ProsodyParseError: <sub> needs alias
'Send it to <language> support.' → ProsodyParseError: <lang> needs xml:lang
'Read the <breaking> news.'      → parsed as Break(700ms); the word vanished from the script
```

The third case is silent corruption. A parser whose entire contract is *leave unrecognised text alone* was deleting a word and putting 700ms of silence where it had been, and reporting nothing. That's the kind of bug that surfaces months later as "the TTS skips words sometimes" and is nearly unfindable from that description.

**How I integrated it:** your `_TAG_NAME_END` lookahead, verbatim, in both `_TAG_RE` and `_STRIP_RE`. My only addition is a parametrised regression test over the seven prefix collisions in the tag set — `<subject>`, `<language>`, `<breaking>`, `<prosodyX>`, `<emphasise>`, `<substitute>`, `<phonemes>` — asserting all three of `has_markup`, the compiled text and `strip_markup`, so this can't quietly come back.

## `source_text` desync — adopted, and I want to record why it's hard to reproduce

I couldn't reproduce this at first, which is worth writing down because the next person won't be able to either. `_coalesce` merges on equal `Attrs`, and a bare `<sub>` run carries default attrs — so the whitespace after it merges into a neighbour and never reaches the plan loop on its own. It needs the substitution nested inside a span that gives it non-default attributes:

```python
'<prosody rate="0.8"><sub alias="ban-DEH-ha">bandeja</sub></prosody> <prosody rate="1.2">rapido</prosody>'

before:  text='ban-DEH-ha ' | source_text='bandeja'     # desynced
after:   text='ban-DEH-ha ' | source_text='bandeja '
```

**How I integrated it:** your `replace()` approach exactly. I added the reachability conditions to the test docstring, because a fix nobody can reproduce is a fix somebody eventually deletes.

## `asyncio.to_thread` — you were right and I'd underweighted this

I'd treated the render loop as I/O-bound because `generate_run` is awaited. It isn't: resample, `trim_edges` and WSOLA are all synchronous numpy, and WSOLA alone is ~0.27s per ten seconds of audio. On a multi-run plan that stalls every coroutine in the process — including the SSE status stream the UI reads *while that exact plan is rendering*. Your UI is what made the cost visible; on a backend-only branch nothing was watching the loop.

**How I integrated it:** `_post_process_run` essentially as you wrote it, plus wrapping `assemble`. One deliberate difference: I did **not** take your `_to_mono` helper, and kept the existing inline `audio.mean(axis=0)` — only because it's used in exactly one place here and extracting it would touch code outside the fix. If you'd rather it be a helper I have no objection, it just wasn't load-bearing for the change.

## Qwen CustomVoice 0.6B — confirmed against the installed package

`qwen_tts/inference/qwen3_tts_model.py:799`:

```python
if self.model.tts_model_size in "0b6": # for 0b6 model, instruct is not supported
    instruct = None
```

So `supports_instruct=True` on that config was simply false, and the compiler was realising `<emphasis>` into a kwarg the library discarded — the exact failure mode the warnings exist to prevent, happening inside the mechanism meant to prevent it.

**How I integrated it:** took the value change, wrote my own comment citing the line. One thing worth flagging for whoever touches this next — that upstream guard is a *substring* test, not equality. It's correct today only because `"1b7" in "0b6"` is False; it would also fire on `"0"`, `"b"`, `"6"` or `""`. Not ours to fix, but it's a thin thread to hang a capability on.

## Where I changed your approach rather than adopting it

Two places, both with reasons rather than preferences.

**Engine capabilities.** Your `_config_for` / `engine_supports_instruct` / `engine_languages` trio is the right observation — capabilities differ *within* an engine, so keying on engine alone can't answer correctly. After your 0.6B fix my per-engine `all(...)` would have denied emphasis to 1.7B, the variant that actually supports it, so this had to change.

I folded the model-size narrowing into the existing `engine_capabilities` rather than adding the three functions, because #1023 already introduces `engine_supports_instruct`/`engine_languages` (without the size parameter) and two competing implementations of the same names across two open PRs would be worse than one temporary one. When #1023 lands, this should collapse into it — and your size parameter is the thing that should survive the merge.

I did change one behaviour: where a requested size doesn't match any config, yours falls back to `configs[0]`; mine stays conservative and reports the `all(...)` answer when no size is given at all. Rationale — a request that hasn't pinned a size can land on any variant, and claiming a capability the chosen variant lacks is the failure we're both trying to avoid, whereas erring the other way only costs a warning that wasn't strictly needed. Genuinely open to being wrong about that one.

**Warnings on the row.** I took your design point wholesale — warnings going only to the server log are, from inside the app, indistinguishable from the model ignoring an instruction. Adopted: the column, the migration, the SSE payload field, `_decode_prosody_warnings` (closely adapted), and the once-per-generation dedupe in `useGenerationProgress`, which is almost your code.

What I changed is where the write happens. Yours persists inline inside `_maybe_render_with_prosody`; mine is an `on_plan` hook on `generate_with_prosody`. The reason is that my streaming path has no `generations` row to write to, and threading a nullable `generation_id` through the shared helper to be ignored half the time reads worse than letting the caller supply what it wants done. Same behaviour otherwise, including firing on trivial plans — which is the case that matters most, since nothing about the audio would hint at what happened.

**Hidden imports** I extended rather than changed: my stack also lists `prosody.annotate`, `prosody.llm_annotate`, `prosody.pipeline`, `services.pronunciation`, `routes.prosody` and `routes.pronunciation`, eleven entries in total.

I have to correct myself on the reasoning, though, because I'd written that the frozen sidecar would fail without them and that isn't true. I checked afterwards by extracting the PYZ from a build that predates both your entries and mine, and all eleven modules were already in it — the route modules are imported at module level, so PyInstaller's analysis reaches the whole subtree on its own. Neither your entries nor mine fix a live failure.

I've kept them anyway, and I think you should too, but for the honest reason rather than the flattering one: the day one of those imports moves inside a function to trim startup time, the graph stops reaching them and it breaks at first use rather than at build time. That's cheap insurance against a plausible refactor. It just isn't a bug fix, and I'd rather tell you that than let you carry a justification I'd already disproved.

## Two things in the branch you'll want to know about

Said plainly because you'd want the same from me, not as a knock on the branch.

**Markup that compiles to a trivial plan gets read aloud.** All three entry points — `stream_speech`, `run_generation`, `generate_audio_sync` — fall through to the original text:

```python
if plan.is_trivial:  →  generate_chunked(tts_model, data.text, ...)   # tags still in it
```

I ran your compiler to find which inputs land there:

```
'<lang xml:lang="en">Hello there</lang>'    is_trivial=True → engine receives the raw tags
'<emphasis level="strong">Hello</emphasis>' is_trivial=True → engine receives the raw tags  (after your emphasis-drop)
'<prosody rate="1.0">Hello there</prosody>' is_trivial=True → engine receives the raw tags
```

The first is ordinary usage — tagging a span in the language you were already generating in. `pipeline.py` passes `plan.nodes[0].text`, the compiled text, which is the only reason I hadn't hit it myself. One-line fix on your side.

**`is_trivial` gaining `source_text is None` is treating the symptom.** It forces respelled plans through the renderer so they miss that fall-through, which works for `<sub>`/`<phoneme>` but leaves the three cases above. On wiring that passes the compiled text it costs something: every respelled sentence gets cut and reassembled for no benefit, which undoes the property that makes respelling the strategy to reach for first — that it doesn't cut the sentence. I've kept my definition, but if the fall-through is fixed I think you'll find you don't need the change either.

**A merge note, not a defect:** your parser drops `_unescape`, so `<sub alias="R&amp;D">` would speak the entity. That's purely because you branched off `feat/prosody-core` and it landed in `feat/prosody-transformer` — nothing to do on your side, just something to watch in the merge.

## Your UI, and what I did to it

The shape you chose is right, and I want to be specific about why rather than just agreeing: a popover of insertable snippets treats the directives as *text the author types*, which is what they are. A toolbar or a rich editor would have implied a document model that doesn't exist here, and would have had to be undone the first time someone pasted a script in.

I've built it out against my backend and credited the branch in each file header:

**`ProsodyHelpPopover`** — substantially yours. Changes: widened 380→400px, added `text`/`language`/`onReplaceText` props, and a slot above the tag list for the draft action below.

I also broke one of your snippets and then fixed it, which is worth admitting because your original was correct. I replaced the literal IPA stress mark in the `<phoneme>` example with `&#712;`, reasoning that keeping the file ASCII was tidier. It isn't: the snippet is a TypeScript string literal rather than JSX text, so nothing decodes it on the way to the textarea, and the parser unescapes the five named XML entities and no numeric references — so the phoneme string would have reached the engine as `ban&#712;dexa` and been spoken as itself. Reverted to your version (`439a1c1`).

**`ProsodyPlanPreview`** — yours, extended to the dictionary. It sends `profile_id` so entries scope to the selected voice, and renders `dictionary_terms`. That last part matters more than it sounds: on your branch the preview compiles hand-written markup, which is the whole story there, but here the pronunciation dictionary injects directives into scripts containing none — so without it, the one case where the author *didn't* type the change is the case the preview stays silent about. Two smaller changes: `total_silence_ms` is computed client-side from the nodes rather than returned by the endpoint, and the empty state keys off `is_trivial && dictionary_terms.length === 0` because my response has no `has_markup` field — I always compile, since the dictionary can make a tagless script non-trivial.

**`ProsodyAnnotateAction`** — new, and the piece a snippet list structurally can't reach. Inserting `<break/>` at the caret helps someone who already knows where the pause goes; it does nothing for someone looking at four paragraphs of narration wondering. This hands the script to the local annotator (`POST /prosody/annotate`, #1036-only, which is why there was nothing in your branch to build against), always shows the suggestion before applying it, and disables itself off the availability probe rather than taking a 409 on click. The backend already refuses a suggestion whose words differ from the input, so the model can fail to help but can't rewrite the script.

## Where that leaves us

Seven commits, on a branch taken from this PR's head so the diff reads against what you already have:

**https://github.com/Lvigentini/voicebox/tree/feat/prosody-user-facing**

- `5b8d8ee` — your four fixes, plus the capability follow-on
- `5e9bd8e` — warnings to the client
- `7bec55b` — the UI
- `439a1c1` — the phoneme snippet, reverted to yours
- `529a64b` — three defects I found reviewing my own integration: the capability fallback above, a stale parse error that could block every subsequent generation once the preview panel was closed, and a redundant DB session in the warnings hook
- `cca69bb` — the packaging correction
- `5877f14` — annotate with whichever Qwen3 is already cached instead of a hardcoded size (my own machine had 0.6B and the feature reported itself unavailable)

Happy to reshape any of this. If you'd rather carry the UI yourself I'll rebase mine out of the way without argument — the four fixes stand on their own and belong in the base PR regardless of who lands the frontend. And if the capability fallback or the `on_plan` split look wrong to you, say so; you've been right about this code four times already.

Lvigentini:
@jamiepine — a question about sequencing rather than a nudge for review, though it's partly that too.

I've been holding back on building UI for this stack on purpose. Your `ui-testing` branch adds `app/src/components/Generation/FloatingGenerateBox.browser.test.tsx` along with the vitest/msw harness, and `FloatingGenerateBox` is exactly where prosody controls have to live — so anything I put there lands on top of tests you've written against the current shape of that component. Specifically, what I've just added collides with three of your five cases:

- two controls in the button row → *renders the generate box wired to the selected profile*
- a pre-submit guard that blocks on unparseable markup → *does not post to /generate when the text is empty* and *posts to /generate on submit*
- a new `prosody_warnings` field on the SSE payload, plus a toast in `useGenerationProgress` → *clears pending state and refetches history when SSE reports completion*, and your msw SSE handler

None of that is hard to reconcile — it's a handful of extra assertions either way — but it's strictly cheaper if we know which side rebases. If `ui-testing` is close to landing I'd rather rebase onto it and write my UI with your harness than have you re-fix five suites after the fact.

The wider version of the same question: I have ten PRs open (#1004, #1005, #1007, #1011, #1012, #1023, #1025, #1026, #1035, #1036), oldest about two weeks, none of them yet looked at by a human. They're a deliberate stack rather than ten unrelated changes — #1035 is the prosody transformer, #1036 builds on it, #1025 is the dictionary #1036 resolves, and #1023 supplies the engine capabilities both of them ask for. Reviewed out of order they'll read as more overlap than there is, and each week they sit they drift further from `main`.

I'm not asking for all ten. What would genuinely help is knowing which one you'd want first, or whether the stack is the wrong shape to be asking you to review at all — I'm happy to collapse it, split it differently, or park most of it and carry a single PR to completion, whichever costs you least. #1011 is a two-line fix that makes the test suite collectable again and would be a cheap place to start.

@hakimio has picked up the prosody work independently and built a UI on it, which is what prompted this — there are now two of us building on an unmerged base, and that's the point where it's worth asking where you want it to go.

Lvigentini:
One more commit on the branch since the comment above — `4d6394b`, a UI for the pronunciation dictionary (#1025), which brings it to eight.

Worth flagging to you specifically, because it lands on the surface you designed and because building it turned up a defect in the plan preview we both have. `ProsodyPlanPreview` replaced its plan rows with *"One pass. The directives changed the words, not how the script is cut"* whenever `is_trivial` was true. On markup-only input that reads fine. With the dictionary it does not: past the empty-state guard, trivial *and* non-empty means the dictionary matched something, and a substitution baked into a single run is precisely the thing an author cannot deduce from their own script. The message was suppressing the only row worth reading. Rows now always render and the note moves underneath.

Two things I confirmed against the live endpoints while wiring it, both relevant to how the strategies read:

- A `phoneme` entry on `qwen` compiles to `<sub alias="...">`, not `<phoneme>`. That is the contract working — `replacement` is required on every entry so the fallback exists — but it means the dictionary is not engine-gated in practice, which is the same property your snippet popover has.
- The compiler's `over_articulated_respelling` warning fires on exactly the respelling style the obvious example invites (`pa-EY-ya`). It is a good warning and it now reaches the user, via the path in `5e9bd8e`.

Still happy to reshape any of it, and the standing offer holds: if you would rather carry the frontend, I will rebase mine out of the way. The four fixes belong in the base PR either way.

PR #1043, opened 2026-08-13
feat(sinks): add Telnyx PSTN sink for outbound call playback

a692570:
# Add Telnyx PSTN sink: call any phone and play generations on the line

## What this adds

Voicebox today runs the full voice I/O loop locally (clone, generate, dictate,
agent speech) but stops at the speaker. There's no way to take a generation and
play it on a real phone call. The roadmap lists "Pipeline routing … webhook +
MCP sinks" and "Platform sinks: Apple Notes, Obsidian, and other opt-in
integrations."

This PR ships the first reference sink, Telnyx, and the convention future sinks
follow. Generation stays local; only the PSTN leg touches the carrier.

## What's in the box

- **`voicebox.call` MCP tool**: dial a number and play a generation on the call.
  Mirrors `voicebox.speak`'s shape (profile resolution, personality rewrite,
  engine override).
- **`POST /call`**: REST wrapper, same pattern as `POST /speak`.
- **`POST /sinks/telnyx/webhook`**: receives Telnyx `call.answered` and
  `playback.ended`. Per-call webhook secret stops third-party triggering.
- **`GET/PUT /sinks/telnyx`**: settings CRUD (api_key masked in responses,
  plus a `missing` array naming any unset required field).
- **`GET /sinks/telnyx/health`**: verify the API key against Telnyx.
- **`TelnyxSettings` and `TelnyxCalls` SQLite tables**: created idempotently on
  first boot via the existing migration pattern.
- **Docs**: new `sinks.mdx` page in the overview.
- **Tests**: `backend/tests/test_telnyx_sink.py`, 17 cases.

## Flow

```
Caller → POST /call (or voicebox.call)
  → Voicebox resolves profile + generation (existing or fresh via /generate)
  → Voicebox POST /v2/calls to Telnyx with connection_id + webhook_url + per-call secret
  → Telnyx returns call_control_id immediately
  → Voicebox persists TelnyxCall row, returns to caller (status="dialing")

Callee picks up → Telnyx POST /sinks/telnyx/webhook (call.answered)
  → Voicebox verifies webhook_secret, claims the row, acks immediately
  → background task waits for the generation to finish (if still "generating")
  → POST /v2/calls/{id}/actions/playback_start with audio_url
  → Telnyx fetches {public_base_url}/audio/{generation_id} and plays it

If auto_hangup → Telnyx POST /sinks/telnyx/webhook (playback.ended)
  → Voicebox POST /v2/calls/{id}/actions/hangup
```

The caller does not block on the callee picking up. The MCP/REST call returns as
soon as Telnyx accepts the dial; playback happens asynchronously via the webhook.

Telnyx wants a 2xx within 2000 ms and retries otherwise, and it queues
consecutive play commands, so the webhook acknowledges before doing any slow
work and claims the call row so a retried delivery can't start a second
playback on the same call.

## HTTP-only, no SDK

Uses `httpx` against `api.telnyx.com/v2`. Three endpoints: `POST /calls`,
`POST /calls/{id}/actions/playback_start`, `POST /calls/{id}/actions/hangup`.
No `telnyx` package, no PyInstaller bundling changes.

`httpx` was already imported at module scope by `services/cloud.py` but was
never declared, so this PR adds it to `requirements.txt`.

## Config

- Settings → Sinks → Telnyx: API key, Call Control Application ID
  (`connection_id`), from-number, `public_base_url` (ngrok/Funnel URL), default
  voice profile, auto-hangup toggle, and an `enabled` master switch.
- Per-sink config, not global env, matching the pattern Apple Notes / Obsidian
  sinks will need.

## Security note on `public_base_url`

Telnyx has to reach this server, but Voicebox binds to 127.0.0.1 and has no
request authentication. A bare `ngrok http 17493` therefore publishes every
route, including the profile, generation, and settings APIs. The docs call this
out and tell users to front the tunnel with a proxy that allows only
`POST /sinks/telnyx/webhook` and `GET /audio/`.

Worth a maintainer opinion: if sinks become a common pattern, Voicebox may want
a first-class "expose these paths only" mode rather than leaving each user to
build it.

## What this PR does NOT do

- No inbound calls (separate sink, separate PR).
- No streaming audio over the call. Uses `playback_start` with an audio URL.
  No SIPREC, no bidirectional audio.
- No Telnyx SDK.
- No move of TTS to the cloud. Generation stays local.
- No frontend UI changes. The backend surface (REST + MCP) is complete; the
  "Call" button on the generation card is a small follow-up PR, and the docs
  say so rather than describing a button that isn't there.
- No Telnyx webhook signature verification yet. See open question 1.

## Files

- `backend/database/models.py`: `TelnyxSettings` + `TelnyxCall` ORM
- `backend/database/__init__.py`: re-exports
- `backend/database/migrations.py`: `_migrate_telnyx_settings`,
  `_migrate_telnyx_calls` (idempotent CREATE TABLE, plus a `connection_id`
  backfill for anyone who ran an earlier build of this branch)
- `backend/models.py`: `CallRequest`, `CallResponse`, `TelnyxSettingsResponse`,
  `TelnyxSettingsUpdate` Pydantic models
- `backend/services/sinks.py`: `TelnyxClient` (thin httpx wrapper) +
  `dial_and_play`, `handle_telnyx_webhook`, `place_call` orchestration
- `backend/routes/sinks.py`: `POST /call`, settings CRUD, webhook receiver
- `backend/routes/__init__.py`: register the new router
- `backend/mcp_server/tools.py`: `voicebox.call` tool, next to `voicebox.speak`
- `backend/tests/test_telnyx_sink.py`: wire contract, config gate, webhook gate
- `requirements.txt`: declare `httpx`
- `docs/content/docs/overview/sinks.mdx`: setup + usage guide
- `docs/content/docs/overview/meta.json`: add the new page

## Why Telnyx

- Carrier-grade PSTN, sub-second call setup, global reach.
- Pay-per-minute, no monthly minimums, fits a hobbyist OSS install.
- REST + webhooks, no SIP stack required on the client.
- A user with a Telnyx account and a $1 number can dial out in a few minutes.

## Setup

1. Sign up at telnyx.com, grab an API key, buy a number (~$1/mo), create a Call
   Control Application, and assign the number to it. Copy the application ID.
2. Expose Voicebox via ngrok, Tailscale Funnel, or Cloudflare Tunnel, ideally
   behind a proxy restricted to the two paths above. Copy the HTTPS URL.
3. Voicebox → Settings → Sinks → Telnyx: paste the key, the application ID, the
   from-number, the public URL, pick a default voice profile, turn on `enabled`.
4. Hit "Test connection".
5. Test via MCP:
   ```python
   await voicebox.call({
     "to": "+15551234567",
     "profile": "Morgan",
     "text": "Deploy complete."
   })
   ```
   Or via curl:
   ```bash
   curl -X POST http://127.0.0.1:17493/call \
     -d '{"to":"+15551234567","profile":"Morgan","text":"Deploy complete."}'
   ```

## Verification

- `backend/tests/test_telnyx_sink.py` passes (17 cases). Covers the request
  bodies sent to Telnyx, the config gate, constant-time token comparison, and
  single-fire playback under retried webhook delivery.
- Unconfigured sink: `POST /call` returns 400 naming the missing fields; the
  MCP tool raises with the same message.
- Existing routes unaffected; the new router is additive.
- Live PSTN verification against a real number is pending on this revision.
  An earlier revision of this branch sent no `connection_id` and used a
  `media_url` field that doesn't exist on `playback_start`, so both the dial
  and the playback would have been rejected by the API. That's fixed here and
  covered by tests, but I want to re-run it end to end on hardware before
  anyone treats the call path as confirmed working.

## Open questions for maintainers

1. **Webhook signature verification.** This PR gates the webhook on a per-call
   `webhook_secret` query param, compared in constant time, failing closed when
   absent. Telnyx also signs every event with `telnyx-signature-ed25519` and
   `telnyx-timestamp` over `{timestamp}|{raw body}`, verifiable against the
   account public key. That authenticates the sender rather than just scoping
   the URL, and it's the better gate for an internet-exposed endpoint, but it
   needs an Ed25519 dependency and a public-key setting. Want it in this PR, or
   as a follow-up?
2. **Sink vs. pipeline-routing roadmap.** This ships a single sink with its own
   settings table, additive, not blocking on the broader pipeline-routing
   refactor. When that lands, the Telnyx sink can be reframed as a sink node
   and the HTTP + webhook logic carries over. Merge the additive version now
   and migrate later, or hold until the sink interface is settled?
3. **Frontend "Call" button.** Backend surface is complete; the React UI is
   left for a follow-up so this PR stays reviewable, and the docs mark it as
   not yet shipped. Acceptable, or should the UI ship here?

## License

MIT, no CLA friction. The contributor is a Telnyx employee but is contributing
as an individual user of the project. No Telnyx affiliation claim beyond the
choice of carrier.

## Summary by CodeRabbit

* **New Features**
  * Added optional Telnyx phone calling with generated voice playback.
  * Added REST and MCP controls for text or existing generations.
  * Added configuration, health checks, call tracking, webhooks, status updates, and automatic hangup support.
  * Added voice profile, language, personality, caller, and masked credential settings.
  * Added validation and retry-safe call event handling.

* **Documentation**
  * Added setup, configuration, usage, security, limitations, and troubleshooting guidance for Telnyx calling.

### DIFF
--- a/backend/database/__init__.py
+++ b/backend/database/__init__.py
@@ -22,6 +22,8 @@ from .models import (
     Project,
     Story,
     StoryItem,
+    TelnyxCall,
+    TelnyxSettings,
     VoiceProfile,
 )
 from .session import engine, SessionLocal, _db_path, init_db, get_db
@@ -44,6 +46,8 @@ __all__ = [
     "Project",
     "Story",
     "StoryItem",
+    "TelnyxCall",
+    "TelnyxSettings",
     "VoiceProfile",
     # Session
     "engine",
--- a/backend/database/migrations.py
+++ b/backend/database/migrations.py
@@ -304,6 +304,12 @@ def _migrate_telnyx_settings(engine, inspector, tables: set[str]) -> None:
     webhook) ship their own singleton tables in their own PRs.
     """
     if "telnyx_settings" in tables:
+        # Table predates connection_id (early builds of the sink omitted it
+        # and every dial 422'd). Backfill the column rather than recreate.
+        if "connection_id" not in _get_columns(inspector, "telnyx_settings"):
+            _add_column(
+                engine, "telnyx_settings", "connection_id VARCHAR", "connection_id"
+            )
         return
 
     with engine.begin() as conn:
@@ -313,6 +319,7 @@ def _migrate_telnyx_settings(engine, inspector, tables: set[str]) -> None:
                 id INTEGER PRIMARY KEY DEFAULT 1,
                 enabled BOOLEAN NOT NULL DEFAULT 0,
                 api_key VARCHAR,
+                connection_id VARCHAR,
                 from_number VARCHAR,
                 public_base_url VARCHAR,
                 default_profile_id VARCHAR REFERENCES profiles(id),
--- a/backend/database/migrations.py
+++ b/backend/database/migrations.py
@@ -43,6 +43,8 @@ def run_migrations(engine) -> None:
     _migrate_generation_versions(engine, inspector, tables)
     _migrate_capture_settings(engine, inspector, tables)
     _migrate_mcp_bindings(engine, inspector, tables)
+    _migrate_telnyx_settings(engine, inspector, tables)
+    _migrate_telnyx_calls(engine, inspector, tables)
     _normalize_storage_paths(engine, tables)
 
 
@@ -290,6 +292,68 @@ def _supports_drop_column(engine) -> bool:
     if engine.dialect.name != "sqlite":
         return True
     return tuple(int(p) for p in sqlite3.sqlite_version.split(".")[:3]) >= (3, 35, 0)
+
+
+def _migrate_telnyx_settings(engine, inspector, tables: set[str]) -> None:
+    """Create the ``telnyx_settings`` table for the Telnyx PSTN sink.
+
+    One singleton row (id=1) holding the user's Telnyx API key, default
+    caller ID, public tunnel URL, and a few toggles. The table itself is
+    created here; the singleton row is lazily inserted by the settings
+    service on first read. Future sinks (Apple Notes, Obsidian, generic
+    webhook) ship their own singleton tables in their own PRs.
+    """
+    if "telnyx_settings" in tables:
+        return
+
+    with engine.begin() as conn:
+        conn.execute(text(
+            """
+            CREATE TABLE telnyx_settings (
+                id INTEGER PRIMARY KEY DEFAULT 1,
+                enabled BOOLEAN NOT NULL DEFAULT 0,
+                api_key VARCHAR,
+                from_number VARCHAR,
+                public_base_url VARCHAR,
+                default_profile_id VARCHAR REFERENCES profiles(id),
+                auto_hangup BOOLEAN NOT NULL DEFAULT 1,
+                updated_at DATETIME
+            )
+            """
+        ))
+    logger.info("Created telnyx_settings table")
+
+
+def _migrate_telnyx_calls(engine, inspector, tables: set[str]) -> None:
+    """Create the ``telnyx_calls`` table that tracks pending outbound calls.
+
+    One row per call. The webhook handler reads this to know which
+    generation to play when Telnyx fires ``call.answered``. The
+    ``webhook_secret`` column is a per-call random token (sent as the
+    ``token`` query param on the webhook URL) so a third party who knows
+    the webhook URL can't trigger playback on someone else's call.
+    """
+    if "telnyx_calls" in tables:
+        return
+
+    with engine.begin() as conn:
+        conn.execute(text(
+            """
+            CREATE TABLE telnyx_calls (
+                call_control_id VARCHAR PRIMARY KEY,
+                generation_id VARCHAR NOT NULL REFERENCES generations(id),
+                to_number VARCHAR NOT NULL,
+                from_number VARCHAR NOT NULL,
+                webhook_secret VARCHAR,
+                status VARCHAR NOT NULL DEFAULT 'initiating',
+                auto_hangup BOOLEAN NOT NULL DEFAULT 1,
+                error TEXT,
+                started_at DATETIME NOT NULL,
+                completed_at DATETIME
+            )
+            """
+        ))
+    logger.info("Created telnyx_calls table")
 
 
 def _normalize_storage_paths(engine, tables: set[str]) -> None:
--- a/backend/database/models.py
+++ b/backend/database/models.py
@@ -317,6 +317,8 @@ class TelnyxSettings(Base):
     id = Column(Integer, primary_key=True, default=1)
     enabled = Column(Boolean, nullable=False, default=False)
     api_key = Column(String, nullable=True)
+    # Call Control Application ID — required on every outbound dial.
+    connection_id = Column(String, nullable=True)
     from_number = Column(String, nullable=True)  # E.164, used as caller ID
     public_base_url = Column(String, nullable=True)  # ngrok/funnel/etc.
     default_profile_id = Column(String, ForeignKey("profiles.id"), nullable=True)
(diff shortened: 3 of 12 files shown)


a692570:
Pushed two commits (`a6c8df0`, `b0021ed`) and rewrote the description. Self-review turned up two bugs that meant the sink couldn't have worked against the live API, so flagging them plainly rather than burying them in a diff.

### Two wire-contract bugs

1. **`POST /v2/calls` was missing `connection_id`.** The Dial API's required set is `[connection_id, to, from]`, so every dial would have been rejected. There was no `connection_id` column, setting, or docs field anywhere in the branch. Added, with a migration that backfills the column for anyone who ran the earlier build.
2. **`playback_start` sent `media_url`.** That field doesn't exist on the action; it's `audio_url` (`media_name` is the separate uploaded-media path). Even on a connected call, playback would have been rejected.

My earlier "tested end to end, playback clean" note in the description was wrong and I've corrected it. Both paths are now covered by tests that assert the exact JSON sent to Telnyx, and I verified those tests fail if either fix is reverted. Live PSTN re-verification is still pending on my side.

### Webhook retry safety

`call.answered` waited up to 30s for the generation while holding the webhook request open. Telnyx wants a 2xx within 2000 ms and retries otherwise, and it queues consecutive play commands, so a slow generation meant a timeout, a retry, and the same clip playing more than once on the call. The old idempotency guard only closed after the wait, so retries sailed past it.

The handler now acks immediately, claims the call row before doing anything slow, and runs the wait in a background task with its own session. Retried deliveries see the claim and exit.

### Smaller fixes

- Telnyx error bodies are unpacked into a 400 instead of `raise_for_status` producing an opaque 500. Misconfiguration is the normal failure here, so the message should say which field Telnyx rejected.
- `/call` now resolves profiles through the same single `resolve_profile` call `/speak` uses. It previously skipped per-client MCP bindings and the global capture default, and ignored the binding's `default_engine` and `default_personality`, despite claiming parity with `voicebox.speak`.
- The `enabled` flag defaulted to false and wasn't documented, so following the setup guide exactly still produced "not configured". The request gate and `dial_and_play` also disagreed about whether to check it. Both now share one `missing_settings()` helper, and `GET /sinks/telnyx` returns a `missing` array so the UI can show what's outstanding.
- Webhook token now compares with `secrets.compare_digest` and fails closed when no secret is stored, instead of `!=` and fail-open.
- Dropped a `call_control_id or call_session_id` fallback that would have silently stored an ID no call-control action accepts.
- `httpx` was already imported at module scope by `services/cloud.py` but never declared; it's in `requirements.txt` now.

### Answers to my own open questions

**1. Webhook signature verification: I'd take it as a follow-up, but I'll do it here if you'd rather.** The per-call token scopes each webhook to one call and fails closed, which is a real gate, but it rides in the query string and so lands in the access logs of whatever proxy fronts the tunnel. Telnyx's Ed25519 signature (`telnyx-signature-ed25519` + `telnyx-timestamp` over `{timestamp}|{raw body}`, checked against the account public key, rejecting timestamps older than five minutes) authenticates the sender rather than just scoping the URL. The cost is a crypto dependency and another settings field, which is why I left it out of a PR that's already large. Your call on the dependency question, since that's a project-level decision.

**2. Sink vs. pipeline routing: I'd merge additive now.** The Telnyx-specific surface is the HTTP client and the two webhook events, and that logic survives a reframing into a sink node unchanged. The settings table is isolated and the routes are namespaced under `/sinks/telnyx`, so the refactor should be a move rather than a rewrite. If you'd rather not carry a second settings-table pattern before the interface exists, I'm happy to hold. The thing I'd want from you either way is a steer on naming so this doesn't become the odd one out later.

**3. Frontend "Call" button: follow-up, and the docs now say so.** The original docs described a button that didn't exist, which was my error; that section is now marked as not yet shipped. Backend surface is complete and usable via MCP and curl today. Happy to add the UI here if you'd prefer to review it as one unit.

### One thing worth your opinion

Telnyx has to reach this server, but Voicebox binds to loopback and has no request auth, so a plain `ngrok http 17493` publishes every route: profiles, generations, settings, deletes. The docs now warn about this and tell users to front the tunnel with a proxy allowing only `POST /sinks/telnyx/webhook` and `GET /audio/`.

That works, but it puts the burden on the user and it's easy to skip. If sinks become a recurring pattern, Voicebox might want a first-class "expose only these paths" mode. That's beyond this PR, and I didn't want to invent that surface unilaterally, but the need shows up the moment any sink needs an inbound callback.

PR #1058, opened 2026-08-21
fix(ui): make long error toasts readable and copyable

hakimio:
A failed generation put the server's error straight into a toast. The transformers "Unrecognized model" error is ~4.8KB, of which the first sentence carries the meaning and the remaining 4.7KB is an alphabetical list of every architecture it knows. In a 420px toast with overflow-hidden and no scroll, that clipped the text at both ends and pushed the close button off-screen: unreadable and undismissable.

- ToastDescription is capped at 40vh and scrolls, wraps on whitespace and breaks long unspaced tokens so a path cannot widen the toast.
- The toast aligns to the start rather than centring, so the title stays visible next to a tall description.
- condenseError() keeps the head of an oversized error, cutting at the first newline or the last sentence end inside a 400-char budget, and reports how many characters it dropped. Short errors pass through untouched.
- When it does truncate, the toast offers a Copy action for the full text and points at Settings -> Logs.

Verified against the real 4795-char error: 4795 -> 400 chars keeping both meaningful sentences.

The hook moves to .tsx to render ToastAction, matching useAutoUpdater.tsx which is a .tsx hook for the same reason. createElement was tried first but this repo's ToastActionElement type is the older shadcn definition (ReactElement<typeof ToastAction>) which only accepts JSX-constructed elements.

------------------------------

## OLD error toast

image

## NEW error toast

image

## Summary by CodeRabbit

* **New Features**
  * Long generation errors are summarized in notifications, with an option to copy the complete error text.
  * Error messages preserve readable formatting and support scrolling for lengthy content.

* **Bug Fixes**
  * Toast content now aligns consistently and wraps long unbroken text more effectively.
  * Lengthy descriptions are constrained to improve readability.
  * Short error messages continue to display as standard error notifications.
  * Clipboard failures now display an error notification.

### DIFF
--- a/app/src/components/ui/toast.tsx
+++ b/app/src/components/ui/toast.tsx
@@ -22,7 +22,9 @@ const ToastViewport = React.forwardRef<
 ToastViewport.displayName = ToastPrimitives.Viewport.displayName;
 
 const toastVariants = cva(
-  'group pointer-events-auto relative flex w-full items-center justify-between space-x-4 overflow-hidden rounded-md border p-6 pr-8 shadow-lg transition-all data-[swipe=cancel]:translate-x-0 data-[swipe=end]:translate-x-[var(--radix-toast-swipe-end-x)] data-[swipe=move]:translate-x-[var(--radix-toast-swipe-move-x)] data-[swipe=move]:transition-none data-[state=open]:animate-in data-[state=closed]:animate-out data-[swipe=end]:animate-out data-[state=closed]:fade-out-80 data-[state=closed]:slide-out-to-right-full data-[state=open]:slide-in-from-top-full data-[state=open]:sm:slide-in-from-bottom-full',
+  // items-start rather than items-center: a long description scrolls inside a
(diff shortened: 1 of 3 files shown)


hakimio:
@coderabbitai

> Add boundary tests for the `CondensedError` contract.

The repo has no frontend test infrastructure at all - no *.test.ts anywhere in app/, web/, or tauri/src, no test script in any package.json, no runner config. The Python side has backend/tests/; the TypeScript side has nothing. And bun run ci is typecheck && build:web, with no test step. 

So, adding a test file here would create a file that nothing ever executes - worse than no test, because it silently rots. Doing it properly means adding a test script and wiring it into ci, which establishes a testing convention for the whole frontend. That's a maintainer's decision, not something to slip into a PR about toast rendering.

ericloyd:
Jesus, yes, please make this go.  I've given up on trying to use this software because I can't get to the error messages to see what needs to be fixed.

Issue #1034, opened 2026-08-10
Feature request: add OmniVoice (k2-fsa) as a TTS engine


edvlos-tech:
Hi! First, thanks for Voicebox — I use it daily.

I'd like to suggest **OmniVoice** (https://github.com/k2-fsa/OmniVoice) as a candidate engine. It's Apache-2.0, zero-shot voice cloning across 600+ languages, and it also supports attribute-based voice design (gender, age, pitch, accent) via an `instruct` string — so it fits both the cloning path and the CustomVoice-style path.

Why it might be a good fit:

- Very broad language coverage, including many that no current Voicebox engine handles
- Simple API shape: `model.generate(text=..., ref_audio=..., ref_text=...)` — nearly identical to Chatterbox, so Pattern B (deferred file paths) applies directly
- Outputs 24 kHz, same as most engines already integrated
- Fast — the authors report RTF as low as 0.025
- Available on PyPI (`pip install omnivoice`)

Issue #791, opened 2026-06-23
Add OmniVoice Model Integration


leoleo-s:
First of all, thank you for the incredible work on this project. Voicebox has become one of the most useful on my workflow.

I would like to suggest adding support for OmniVoice as a new TTS model.

OmniVoice has recently gained significant attention due to its high-quality speech synthesis and competitive performance compared to many existing open-source and commercial alternatives and I find It's much faster than Qwen TTS already available on Voicebox.

For me, this would turn Voicebox the ultimate TTS software out there.

Thank you for considering this request. I believe OmniVoice would be a great addition to the Voicebox ecosystem and would be appreciated by a large portion of the community.

Best regards.

Issue #526, opened 2026-04-21
HF Hub Offline


izidre:
Hello, I get an error when selecting the custom voice "qwen".

`
2026-04-21 10:14:45,296 - backend.backends.qwen_custom_voice_backend - INFO - Loading Qwen CustomVoice 1.7B on cuda...
2026-04-21 10:14:45,297 - backend.utils.hf_offline_patch - INFO - [offline-guard] qwen-custom-voice-1.7B is cached � forcing offline mode
`torch_dtype` is deprecated! Use `dtype` instead!
2026-04-21 10:14:45,297 - qwen_tts.core.models.configuration_qwen3_tts - INFO - speaker_encoder_config is None. Initializing talker model with default values
2026-04-21 10:14:45,298 - qwen_tts.core.models.configuration_qwen3_tts - INFO - talker_config is None. Initializing talker model with default values
2026-04-21 10:14:45,298 - qwen_tts.core.models.configuration_qwen3_tts - INFO - speaker_encoder_config is None. Initializing talker model with default values
2026-04-21 10:14:45,299 - qwen_tts.core.models.configuration_qwen3_tts - INFO - code_predictor_config is None. Initializing code_predictor model with default values
2026-04-21 10:14:45,299 - qwen_tts.core.models.configuration_qwen3_tts - INFO - code_predictor_config is None. Initializing code_predictor model with default values
2026-04-21 10:14:46,863 - qwen_tts.core.tokenizer_12hz.configuration_qwen3_tts_tokenizer_v2 - INFO - encoder_config is None. Initializing encoder with default values
2026-04-21 10:14:46,863 - qwen_tts.core.tokenizer_12hz.configuration_qwen3_tts_tokenizer_v2 - INFO - decoder_config is None. Initializing decoder with default values
2026-04-21 10:14:47,386 - backend.utils.progress - ERROR - Marked qwen-custom-voice-1.7B as error: Cannot reach https://huggingface.co/api/models/Qwen/Qwen3-TTS-12Hz-1.7B-CustomVoice: offline mode is enabled. To disable it, please unset the `HF_HUB_OFFLINE` environment variable.
Traceback (most recent call last):
  File "backend\services\generation.py", line 64, in run_generation
  File "backend\backends\__init__.py", line 400, in load_engine_model
  File "backend\backends\qwen_custom_voice_backend.py", line 93, in load_model_async
  File "asyncio\threads.py", line 25, in to_thread
[...]

Issue #517, opened 2026-04-20
Model Request: Add OmniVoice


madkrystiank:
Add OmniVoice model

PR #1062, opened 2026-08-22
feat(engines): add OmniVoice TTS engine (23 languages)

### LINKED ISSUE #380: support https://zhu-han.github.io/omnivoice/
pls add https://zhu-han.github.io/omnivoice/

pippo73:
Adds [OmniVoice](https://github.com/k2-fsa/OmniVoice) (k2-fsa) as an eighth TTS engine: zero-shot cloning across 600+ languages, attribute-based voice design, 24 kHz output, Apache 2.0.

Closes #380. Also covers the duplicate requests in #388, #389, #517, #791 and #1034.

`PROJECT_STATUS.md` lists OmniVoice as `Candidate (#380)` with its comparison row still empty, so this PR also fills in what the evaluation would have found.

## Why nobody has opened this PR before

The engine was never the problem. `transformers` was.

OmniVoice needs `transformers >= 5.3` for its audio codec, `HiggsAudioV2TokenizerModel`. We are capped at `<= 4.57.6`, and not loosely: `backend/utils/hf_offline_patch.py` wraps `PreTrainedTokenizerBase._patch_mistral_regex` (#526, three releases to settle) and `backend/pyi_rth_torch_compiler_disable.py` rewrites the source of `transformers.masking_utils` to force `_is_torch_greater_or_equal_than_2_6 = False`. Both read 4.57.x internals. Moving to 5.x to gain one engine would break the seven already shipping.

So OmniVoice is installed `--no-deps` and the gap is bridged locally — the same treatment `mlx-lm` and `mlx-audio` already get in the justfile, for the same reason.

The weights for the codec already ship inside the OmniVoice repo (`audio_tokenizer/model.safetensors`). Only the modelling code was missing, so that is what gets vendored.

## What is in `backend/vendor/`

- **`higgs_audio_v2_tokenizer/`** — the codec's two source files from `transformers`, imports rewritten, bodies verbatim. `scripts/sync-higgs-tokenizer.py` regenerates them against any upstream ref and fails loudly if a relative import survives the rewrite, so a re-sync cannot silently produce a module that resolves against the wrong package.
- **`_compat.py`** — what 4.57.6 does not have. Three of these are one-liners (`conv1d_output_length`, the `initialization` module, no-op `auto_docstring`/`strict`). The fourth is not: transformers 5 made configs **dataclasses**, and the vendored config relies on `__post_init__` to promote nested dicts into real sub-config objects. 4.57.6 never calls it, which leaves `acoustic_model_config` as a raw dict and fails later inside `to_diff_dict()`. `_compat.PreTrainedConfig` reproduces just that part of the 5.x protocol.
- **`backend/utils/transformers5_compat.py`** — grafts the codec onto the `transformers` namespace and teaches 4.57.6 to accept the list form of `extra_special_tokens` that OmniVoice's `tokenizer_config.json` writes (4.57.6 calls `.keys()` on it and dies).

One trap worth recording: the first import of `transformers.processing_utils` re-runs `transformers/__init__.py` and rebinds `sys.modules["transformers"]` to a fresh `_LazyModule`. Any `setattr` performed before that is silently lost. The shim therefore installs *after* importing the vendored package, not before.

## Two bugs found along the way

**A ~0.8 GB VRAM leak.** `HiggsAudioV2TokenizerPreTrainedModel._get_conv1d_layers` carries an `@lru_cache` on an *instance method*. The cache key includes `self` and the cache lives on the function object, so the first call pins the tokenizer for the life of the process. Encoding one voice prompt is enough, after which `unload_model()` stranded 0.82 GB. This is an upstream `transformers` bug; I left the vendored file verbatim and clear the cache explicitly on unload, so a re-sync cannot drop the fix. Unload now goes 2.04 GB → 0.01 GB.

**`CLONING_ENGINES` in `backend/services/profiles.py`.** This allow-list is separate from the frontend's and is not mentioned in `docs/content/docs/developer/tts-engines.mdx` — Phase 2 of that guide says the service layer needs no changes. Without adding `omnivoice` to it, every cloned-profile generation is rejected at the route with `Engine 'omnivoice' does not support cloned voice profiles`. Worth adding to the guide; happy to do that in a follow-up. Note that the Python-level test passed the whole time, because it calls the backend directly and bypasses the service.

## Engine behaviour worth knowing

`instruct` here is a **closed vocabulary** — gender, age, pitch, whisper, English accent, Chinese dialect — not the free-form prose Qwen CustomVoice accepts. Out-of-vocabulary values raise upstream, so the backend rewraps that into an error naming the categories. Worth reflecting in the UI hint text eventually.

Languages are limited to the 23 codes already in the language enum, with `ar` mapped to `arb` because OmniVoice enumerates Arabic varieties rather than the macrolanguage. Exposing more of the 600+ means widening `ALL_LANGUAGES`, the request regex and the frontend; that felt like a separate change.

## Verification

Verified on Linux / CUDA (RTX 5060 Ti, 16 GB), `transformers 4.57.6`, `torch 2.11.0+cu128`, Python 3.13.

Because "the array has a plausible shape" is not evidence of intelligible speech, every generated clip was transcribed back with Whisper large-v3-turbo and scored against the input text:

| Case | WER |
|---|---|
| Auto voice (EN) | 0.00 |
| Voice design, `female, low pitch, british accent` (EN) | 0.00 |
| Cloning (EN) | 0.00 |
| Cloning (IT) | 0.10 — apostrophe normalisation only |
| Cloning (AR) | 0.33 |
| Through `POST /generate` | transcript identical |

On the Arabic sample I should be honest: 0.33 on a three-word sentence is one word rendered differently, which is too short to conclude anything about either OmniVoice or Whisper. It deserves a longer test.

Also covered: model load, voice-prompt cache round-trip through `torch.load(weights_only=True)` (which is why the prompt is stored as a plain dict rather than the `VoiceClonePrompt` dataclass), unload, `POST /generate` end to end against a cloned profile, and the engine appearing in the built UI.

Adds `backend/tests/test_omnivoice_compat.py` — 25 tests covering the compat layer and the engine wiring, no model download, no GPU. The existing suite passes unchanged (`test_progress.py::test_hf_progress_tracker` fails both with and without this branch).

RTF measured 0.17–0.19 including Python overhead, on a card that was not idle. Upstream claims 0.025 under batch conditions.

## Not verified

**The PyInstaller directives in `build_binary.py` have never been exercised.** They are written from the Phase 0 audit, not from a build. I have no Windows or macOS machine for this, and there are no Linux binaries to test against.

The audit is at least encouraging: OmniVoice has no `inspect.getsource`, no `@typechecked`, no `@torch.jit.script`, and no data files resolved from system paths — the opposite profile to LuxTTS or TADA. The directives are mostly hidden-imports for `AutoModel.from_config` resolution plus `--copy-metadata omnivoice`. But please treat that section as needing a real `just build` before release.

## About the second commit

`fix(deps): make numba installable on Python 3.13` is independent of OmniVoice — `requirements.txt` is currently uninstallable on 3.13 because numba 0.60.x has no cp313 wheel. It keeps the `<0.61` cap from 38bf96f for `python_version < "3.13"`, where the CI wheel reason still applies, and relaxes it only above. Nothing changes for CI or current release builds. Happy to split it into its own PR if you would rather review it separately.

## Summary by CodeRabbit

* **New Features**
  * Added OmniVoice as a text-to-speech engine with support for 27 languages.
  * Added zero-shot voice cloning, voice design, and cloned voice profile support.
  * Added OmniVoice model management and reference-audio processing.
  * Added deterministic generation and instruction-based controls.

* **Bug Fixes**
  * Improved compatibility with supported Transformers versions and tokenizer behavior.

* **Tests**
  * Added coverage for OmniVoice functionality, caching, language handling, and concurrent loading.


pippo73:
Pushed 166588f addressing the automated review. I checked each finding against the code before acting on it; all five were real, and one of them was the most useful thing anyone could have told me about this PR.

## Accepted

**Voice design was unreachable from the UI.** `supportsInstruct` in `useGenerationForm.ts` and both JSX gates in `FloatingGenerateBox.tsx` tested `engine === 'qwen_custom_voice'` as a literal, so OmniVoice sent `instruct: undefined` and the control never rendered. Half of what makes this engine interesting was dead on arrival. The predicate now lives once, in `app/src/lib/constants/engines.ts` — three hand-written copies is precisely why adding an engine meant remembering three places.

**`load_model` had no concurrency guard.** Both `create_voice_prompt()` and `generate()` call it, so two requests arriving together could clear the `self.model is not None` check before either assigned it. At 3.3 GB a duplicate load is an OOM, not just wasted work. Now uses the double-checked `asyncio.Lock` from `chatterbox_backend.py`.

Worth flagging separately: `luxtts_backend.py` and `kokoro_backend.py` still have the unguarded form. Same shape, smaller models, so it would show up as wasted load time rather than an OOM — but it looks like the same latent bug, and I left them alone rather than widen this PR.

**The voice-prompt cache ignored `format_version` on hits.** Only `_rebuild_clone_prompt` checked it, at generation time, so a future version bump would turn every affected profile into a hard failure while the stale entry stayed cached. Now checked at lookup, where a mismatch is simply a miss and the prompt gets re-encoded.

**`--hidden-import` was not enough for the codec sub-models.** `AutoModel.from_config()` resolves `modeling_dac` and `modeling_hubert` lazily and the parent package alone does not pull them in. Switched to `--collect-submodules`. Still unverified, like the rest of the bundling work.

**`torch>=2.2.0` was too low** — the vendored codec uses `torch.nn.Buffer`.

One correction here: the review says 2.4.0, but that is not right. Checking the PyTorch tags, `class Buffer` appears in `torch/nn/parameter.py` only from **v2.5.0** — it is absent in both v2.3.0 and v2.4.0. So the floor is 2.5.0, which is what I used.

This one raises the minimum for the whole project rather than just this engine, so it is your call. The old floor was already well below what the CUDA/ROCm install paths and the transformers 4.57 workarounds actually run against, but if you would rather keep `>=2.2` and have OmniVoice fail loudly on older torch, say so and I will drop it.

## Declined

The docstring-coverage check (32% against an 80% threshold). It is computed over every function in the diff, and most of those are the ~650 lines vendored verbatim from `transformers`. Documenting them would defeat the point of keeping the file re-syncable.

## Also in this push

The fixture teardown in the new tests was broken in a way the review understated. `_set_model_specific_special_tokens` is defined on `SpecialTokensMixin`, not on `PreTrainedTokenizerBase`, so the saved value was *always* `None` and the test double survived for the rest of the pytest session, silently shadowing the real method for any later tokenizer test. Two regression tests now cover the teardown itself.

9a71317 is a pure line-ending normalization — I had committed `test_omnivoice_compat.py` with CRLF from a Windows host. `git diff -w` against its parent is empty. It is isolated so the review commit stays readable at 91 lines instead of 570.

## Follow-up worth considering

`ModelConfig.supports_instruct` already exists on the backend and is read by nothing — I grepped, only my new test touches it. That is the structural reason the frontend carried a hardcoded engine list at all. If it were surfaced through `/models/status`, the UI could derive the capability instead of duplicating it, and this class of bug would stop being possible. Happy to do that in a separate PR if you want it.

Same for the guide: `docs/content/docs/developer/tts-engines.mdx` says Phase 2 needs no service-layer changes, but `CLONING_ENGINES` in `backend/services/profiles.py` has to be updated for any cloning engine. That cost me a debugging round, and the Python-level test never caught it because it calls the backend directly and bypasses the service. Worth a line in the doc.

167 backend tests pass, typecheck clean. The PyInstaller section still needs a real `just build` on a platform I do not have.

pippo73:
Pushed a62b65a. One finding this round, and it was right.

**`_ensure_asr()` had no lock.** It is called from `_encode_sync`, which runs in the worker thread `asyncio.to_thread` spawns, so two clone requests without reference text land in separate threads and can both clear the `self._asr_loaded` check before either sets it. Whisper large-v3-turbo is ~1.6 GB, so this is the same class of waste the model load lock from 166588f exists to prevent.

`threading.Lock` rather than `asyncio.Lock` — this path does not run on the event loop, so the asyncio primitive would be the wrong one. The review had that right too.

The regression test reproduces the race rather than mocking it: eight threads meet at a `threading.Barrier` and the fake `load_asr_model` sleeps, so every worker gets past an unguarded check. With the lock removed, `assert len(calls) == 1` fails as `assert 8 == 1` — the test catches the bug, it does not merely accompany the fix.

169 backend tests pass, up from 167. `test_progress.py::test_hf_progress_tracker` still fails, as it does on main without this branch.

That is everything from both review rounds addressed except the docstring-coverage warning, which I declined earlier for the reason given above. The PyInstaller directives remain the one part of this PR I cannot verify — they need a real `just build` on a platform I do not have.

pippo73:
Took the title check, and pushed fefc8c5 for the same problem in the UI.

The check was right: the title said 600+ languages, the PR wires 23. Retitled to `feat(engines): add OmniVoice TTS engine (23 languages)`.

Chasing that claim turned up two more places the user actually reads, not just the title:

- `display_name` in the model registry, the label in the model dropdown. Now `OmniVoice (Multilingual)`, matching `Chatterbox TTS (Multilingual)` — the engine that covers the same 23 codes.
- The engine picker subtitle, which sits directly above the language dropdown where a reader can count the entries. Now `23 languages, voice design`, the form the chatterbox entry two lines up already uses.
- The model description in server settings, the one place with room to state both facts rather than pick one: the model covers 600+, Voicebox exposes 23.

Fixing only the title would have left `600+ languages` sitting on top of a list of 23, which is where the claim is most visible.

The comments in `languages.ts`, `omnivoice_backend.py` and the registry already explained the gap and are unchanged. Happy to widen the language enum instead if you would rather have the coverage than the honest label, but that changes the API contract, so it seemed like your call rather than mine.

## Correction to my earlier comment

I said the docstring coverage was dominated by the vendored transformers code. That was overstated. Counting them: 42 of the 117 analyzed functions are in `backend/vendor/higgs_audio_v2_tokenizer/` — about a third, not most. Those 42 genuinely cannot take docstrings, since `scripts/sync-higgs-tokenizer.py` regenerates the files from upstream and hand-edits would both be wiped and make it impossible to tell at a glance whether a file is still faithful to the original. For the other 75 the argument does not apply. Still not something I would treat as blocking, but it is a narrower claim than the one I made.

169 backend tests pass, tsc clean. The two biome warnings in the touched frontend files predate this branch — checked with `git stash`.

Alexander3847575:
Could we get a temporary nightly build while we hope for a merge? This would be the best GUI for OmniVoice on Mac by a long shot! (DotsTTS would also be great, I may get around to it if no one else does)

PR #1074, opened 2026-08-27
fix(backend): release memory when unloading MLX models

JnyRoad:
## Summary
On Apple Silicon (the default MLX backend), clicking "Unload" on a loaded model did not actually shrink the process's memory footprint. `unload_model()` only dropped the Python reference to the model (`del self.model`), but MLX keeps freed array buffers in its own allocator pool for reuse instead of returning them to the OS — the memory stayed resident until the process exited. Separately, the voice-clone prompt cache is a process-lifetime dict that nothing ever cleared on model unload, so previously-generated prompts also stayed referenced forever.

## Changes
- Add `empty_mlx_cache()` (`backend/backends/base.py`), wrapping `mx.clear_cache()`, and call it from all three MLX `unload_model()` implementations: `MLXTTSBackend`, `MLXSTTBackend`, `MLXQwenLLMBackend`.
- Add `clear_voice_prompt_memory_cache()` (`backend/utils/cache.py`) — clears the in-memory voice-prompt cache only, leaving the on-disk `.prompt` cache intact so a later generation still reloads from disk instead of recomputing. Wire it into every TTS unload path (`services/tts.py::unload_tts_model`, and the `qwen_custom_voice` / generic branches of `unload_model_by_config`). Whisper and the LLM backends never produce voice prompts, so their unload paths are left untouched.

## Testing
- New unit tests: `backend/tests/test_mlx_unload_clears_cache.py`, `backend/tests/test_voice_prompt_cache_unload.py` — 8 tests, all passing.
- Verified end-to-end on Apple Silicon against real cached models (Qwen TTS 1.7B, Whisper Turbo, Qwen3 0.6B): loaded each through the running app, unloaded via the real `/models/{name}/unload` endpoint, and confirmed via `mx.get_cache_memory()` / `mx.get_active_memory()` that the MLX allocator's cache drops to 0 on every load/unload cycle (repeated 3x for the LLM path to rule out cumulative growth).
- Ran a real voice-clone generation end to end (synthetic reference audio, real profile, real `/generate` call) and confirmed the in-memory prompt cache goes from 1 entry to 0 on unload, while the on-disk `.prompt` file is left intact.

## Risks
No behavior change for non-MLX (CUDA/CPU) backends. The voice-prompt memory cache is cleared on any TTS engine's unload rather than tracked per-engine (the cache key isn't currently engine-tagged), so if multiple TTS engines are loaded simultaneously, unloading one will also drop the in-memory cache for prompts created by another still-loaded engine — they'll transparently reload from the on-disk cache on next use.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved memory cleanup when unloading speech synthesis, speech recognition, and language models.
  * Voice prompt data is now cleared from memory after text-to-speech models are unloaded, while saved disk cache files remain intact.
  * Prevented unnecessary cache clearing when unloading models that are already inactive.
  * Reduced retained memory after MLX-powered models are unloaded, helping support more reliable switching between models.

* **Tests**
  * Added regression coverage for MLX memory cleanup and voice prompt cache behavior.

### DIFF
--- a/backend/backends/__init__.py
+++ b/backend/backends/__init__.py
@@ -566,6 +566,7 @@ def unload_model_by_config(config: ModelConfig) -> bool:
     """Unload a model given its config. Returns True if it was loaded, False otherwise."""
     from . import get_tts_backend_for_engine
     from ..services import tts, transcribe, llm as llm_service
+    from ..utils.cache import clear_voice_prompt_memory_cache
 
     if config.engine == "whisper":
         whisper_model = transcribe.get_whisper_model()
@@ -595,6 +596,7 @@ def unload_model_by_config(config: ModelConfig) -> bool:
         loaded_size = getattr(backend, "_current_model_size", None) or getattr(backend, "model_size", None)
         if backend.is_loaded() and loaded_size == config.model_size:
             backend.unload_model()
+            clear_voice_prompt_memory_cache()
             return True
         return False
 
@@ -602,6 +604,7 @@ def unload_model_by_config(config: ModelConfig) -> bool:
     backend = get_tts_backend_for_engine(config.engine)
     if backend.is_loaded():
         backend.unload_model()
+        clear_voice_prompt_memory_cache()
         return True
     return False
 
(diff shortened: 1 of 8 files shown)


JnyRoad:
@coderabbitai full review

PR #1078, opened 2026-08-30
fix(qwen): guard all backends against EOS-miss runaway audio

### LINKED ISSUE #1077: Qwen TTS (PyTorch/CPU): missed codec EOS appends minutes of gibberish to long generations — runaway retry from #964 only covers MLX
## Symptoms

Long Qwen TTS generations (voice-clone, zh) play normal speech for the first chunk, then degrade into minutes of unintelligible gibberish / low-level murmur until the end of the file. Status still shows "completed".

## Forensics (local repro, Windows, CPU, v0.5.0)

Analyzed 3 affected long generations from my own history (`generations/*.wav` + `voicebox.db`), plus healthy short ones as controls.

**Signal analysis** — per-10s windows of RMS / zero-crossing rate / spectral flatness:

| generation | text | duration | onset of corruption |
|---|---|---|---|
| `fd38149e` (0.6B) | 1719 chars | 973 s | ~360 s |
| `402c4a22` (1.7B) | 1719 chars | 846 s | ~270 s |
| `f84cdd7d` (1.7B) | 1835 chars | 880 s | ~280 s |

The onsets align with chunk boundaries (800-char chunks). Healthy speech has ZCR ≈ 0.05–0.09; corrupted segments jump to ZCR ≈ 0.13–0.28, and the 0.6B clip drops ~25 dB into low-level murmur.

**Whisper large-v3-turbo transcription** of sampled windows:

- normal segments transcribe verbatim to the source text (e.g. at 805 s: "离开我们已经很久了…一个劳动者可以被尊重的时代…");
- corrupted segments only produce hallucinations ("so", "Thank you.") — the audio carries no recoverable content.

**The smoking gun**: `fd38149e` chunk 3 is 175 chars (~40 s of normal speech), yet the runaway segment runs ~613 s. At 12 Hz that is almost exactly the checkpoint's `max_new_tokens: 8192` (683 s). The model finished (or derailed at) the chunk and never emitted EOS — it decoded to the hard cap.

Same signature as QwenLM/Qwen3-TTS#118 (~0.5% of inferences), and sgl-project/sglang-omni#1179 measured the mechanism: during the failure state the EOS probability collapses to ~1e-14 (rank ~1500+, far outside `top_k=50`), so sampling can never pick it. It is **model-intrinsic, not backend-specific** — which is why it shows up on PyTorch CPU too, not just MLX.

## Why the existing guard misses it

1. #964 added `has_tts_runaway` + chunk-split retry, but enabled it with `retries_runaway = backend_type == "mlx"` — the PyTorch CPU/GPU path ships unprotected. Every Windows/Linux user is in the blind spot.
2. `has_tts_runaway` only flags "speech → long internal silence → speech". The CPU runaway shapes I measured are (a) continuous low-level gibberish and (b) mid-chunk degradation into ZCR-abnormal babble at normal level — neither contains the internal-silence signature.
3. The checkpoint's `max_new_tokens: 8192` is used as-is; nothing bounds the decode to what the text can plausibly fill.
4. The desktop app always sends `max_chunk_chars` (factory default 800, ≈ 200–300 s of audio per autoregressive roll). Long rolls raise the odds of an EOS miss and enlarge the blast radius; engines have no opinion of their own.

## Proposed fix (PR to follow)

1. Enable the runaway retry for `qwen` / `qwen_custom_voice` on **all** backends (EOS misses are model-intrinsic).
2. Add a duration-based detector (`output far longer than its text can plausibly fill`) alongside the internal-silence detector — this catches the 613-s-of-gibberish shape.
3. Bound each Qwen decode with a per-call `max_new_tokens` budget (12 tokens/char + floor, model cap as ceiling). Normal delivery never comes close (~6 tokens/char at the slowest sane pace); a missed EOS now appends at most a fraction of the previous damage.
4. Give Qwen engines a 300-char default chunk (≈ 60–100 s rolls) used when the request doesn't set one; the app stops sending the factory 800 unless the user actually customized the slider.

Happy to send the PR if this sounds reasonable.

**Environment**: Voicebox v0.5.0, Windows, `backend_variant=cpu`, Qwen 0.6B/1.7B Base, voice-clone profiles, zh.

zhongerdeajie:
## Summary

- enable the existing runaway retry (#964) for `qwen` / `qwen_custom_voice` on **all** backends — the guard was MLX-only, leaving the PyTorch CPU/GPU path unprotected against the same EOS-miss failure
- add a duration-based runaway detector (`exceeds_plausible_speech_duration`, non-whitespace chars vs audio length) alongside the internal-silence detector; combined as `detect_tts_runaway(audio, sample_rate, chunk_text)`
- bound every Qwen decode with a per-call `max_new_tokens` budget (`estimate_max_new_tokens`: 12 tokens/char + floor 1500, model cap 8192)
- give Qwen engines a 300-char default chunk, used when the request doesn't set one; the app omits the field while the chunk slider sits on its factory value

## Root cause

Qwen3-TTS occasionally misses its codec EOS. During the failure state the EOS probability collapses to ~1e-14 (rank ~1500+, far outside `top_k=50`), so sampling can never terminate the decode — measured model-intrinsic on PyTorch CPU, MLX and sglang (sgl-project/sglang-omni#1179; QwenLM/Qwen3-TTS#118). The checkpoint ships `max_new_tokens: 8192` ≈ 11.4 minutes of audio, so one miss can append many minutes of gibberish that still gets saved as a `completed` generation.

Two things made the existing protection miss it in practice:

1. `retries_runaway = backend_type == "mlx"` — PyTorch path had no retry at all.
2. `has_tts_runaway` only flags "speech → ≥2s internal silence → speech". The CPU runaway shapes are continuous low-level gibberish or mid-chunk babble at normal level — neither contains a silence gap.

## Reproduction and validation

Signal analysis (RMS / zero-crossing rate / spectral flatness per 10 s) + Whisper large-v3-turbo transcription of 3 affected local generations (details in #1077):

| generation | text | duration | corruption onset | runaway tail |
|---|---|---|---|---|
| `fd38149e` (0.6B) | 1719 chars | 973 s | ~360 s | ~613 s ≈ the 8192-token cap (683 s) |
| `402c4a22` (1.7B) | 1719 chars | 846 s | ~270 s | ~520 s |
| `f84cdd7d` (1.7B) | 1835 chars | 880 s | ~280 s | ~600 s |

- corrupted segments: ZCR 0.13–0.28 vs 0.05–0.09 for healthy speech; Whisper transcribes them as empty hallucinations ("so", "Thank you.") while healthy segments transcribe verbatim to the source text
- `fd38149e` chunk 3 is 175 chars (~40 s of normal speech) yet ran 613 s — the decode hit the token cap without EOS

CPU smoke on 0.6B Base (voice clone, zh, 44-char input):

- healthy run: 11.8 s audio against the 125 s budget — not truncated, detectors negative
- a second run happened to hit a real EOS miss (44 chars → 119.9 s of murmur): the budget cut the tail from 682 s to 125 s and the duration detector flagged it — both layers behaving as intended

## Tests

- `pytest backend/tests/test_qwen_runaway_retry.py backend/tests/test_qwen_generation_guards.py -q` — 21 passed (budget math, duration detector incl. whitespace-padding regression, engine wiring on both backends, retry recovery + persistent-failure)
- full backend suite minus pre-existing failures unrelated to this change — 155 passed
- `bun run typecheck` — clean
- ruff clean on all lines added by this PR

Fixes #1077

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Qwen speech generation reliability by limiting excessive decoding and detecting unusually long or runaway audio.
  * Applied engine-specific chunking defaults when the chunk-size setting remains unchanged.
  * Enabled safer retry behavior for affected Qwen voice-generation modes while preserving other engines’ behavior.
  * Improved duration checks so whitespace padding cannot mask runaway audio.
* **Tests**
  * Added coverage for decode limits, runaway detection, retry behavior, chunk defaults, normal audio duration, and trailing silence.

### DIFF
--- a/app/src/lib/hooks/useGenerationForm.ts
+++ b/app/src/lib/hooks/useGenerationForm.ts
@@ -45,7 +45,16 @@ export function useGenerationForm(options: UseGenerationFormOptions = {}) {
   const generation = useGeneration();
   const addPendingGeneration = useGenerationStore((state) => state.addPendingGeneration);
   const { settings: genSettings } = useGenerationSettings();
-  const maxChunkChars = genSettings?.max_chunk_chars ?? 800;
+  // Factory default for the chunk slider. When the user hasn't customized
+  // it, leave max_chunk_chars unset so the backend can pick the engine's
+  // own default — Qwen engines use a smaller chunk as part of their
+  // EOS-miss runaway guard (300 chars vs the global 800).
+  const FACTORY_DEFAULT_CHUNK_CHARS = 800;
+  const customChunkChars =
+    genSettings?.max_chunk_chars != null &&
+    genSettings.max_chunk_chars !== FACTORY_DEFAULT_CHUNK_CHARS
+      ? genSettings.max_chunk_chars
+      : undefined;
   const crossfadeMs = genSettings?.crossfade_ms ?? 50;
   const normalizeAudio = genSettings?.normalize_audio ?? true;
   const selectedEngine = useUIStore((state) => state.selectedEngine);
@@ -153,7 +162,7 @@ export function useGenerationForm(options: UseGenerationFormOptions = {}) {
         engine,
         instruct: supportsInstruct ? data.instruct || undefined : undefined,
         personality: data.personality || undefined,
-        max_chunk_chars: maxChunkChars,
+        max_chunk_chars: customChunkChars,
         crossfade_ms: crossfadeMs,
         normalize: normalizeAudio,
         effects_chain: effectsChain?.length ? effectsChain : undefined,
--- a/backend/backends/__init__.py
+++ b/backend/backends/__init__.py
@@ -57,6 +57,10 @@ class ModelConfig:
     size_mb: int = 0
     needs_trim: bool = False
     retries_runaway: bool = False
+    # Engine-specific chunking default. Used when a generation request does
+    # not set max_chunk_chars explicitly. None means "use the global
+    # DEFAULT_MAX_CHUNK_CHARS".
+    default_chunk_chars: Optional[int] = None
     supports_instruct: bool = False
     languages: list[str] = field(default_factory=lambda: ["en"])
 
@@ -233,9 +237,18 @@ def _get_qwen_model_configs() -> list[ModelConfig]:
         repo_1_7b = "Qwen/Qwen3-TTS-12Hz-1.7B-Base"
         repo_0_6b = "Qwen/Qwen3-TTS-12Hz-0.6B-Base"
 
-    # mlx-audio can continue after an EOS miss with silence followed by
-    # codec noise. Retry only the affected text as smaller chunks.
-    retries_runaway = backend_type == "mlx"
+    # Qwen3-TTS can miss its codec EOS on ANY inference path — the EOS
+    # probability collapses to ~1e-14 (rank ~1500+, far outside top_k=50)
+    # and the model keeps decoding until max_new_tokens. This is
+    # model-intrinsic (measured on PyTorch CPU, MLX and sglang; see
+    # QwenLM/Qwen3-TTS#118 and sgl-project/sglang-omni#1179), so the
+    # retry guard must not be limited to the MLX backend.
+    retries_runaway = True
+
+    # Smaller chunks keep each autoregressive roll short, which both
+    # lowers the odds of an EOS miss and caps the damage when one
+    # happens. 300 chars ≈ 60-100s of audio per roll.
+    default_chunk_chars = 300
 
     return [
         ModelConfig(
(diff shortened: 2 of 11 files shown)


zhongerdeajie:
@coderabbitai Thanks for the review — pushed 7b23da9 addressing both points:

1. **Duration guard whitespace**: `exceeds_plausible_speech_duration` now counts non-whitespace characters only (`len("".join(text.split()))`), so blank lines and padding can't raise the plausibility threshold. Added a regression test (`test_duration_detector_ignores_whitespace_padding`) where 100 spoken chars + 1500 padding chars still flags a 110 s clip.

   Note the deliberate asymmetry with `estimate_max_new_tokens`: it still uses raw `len(text)`, because there whitespace errs on the safe side — a slightly larger decode budget can never truncate healthy speech, while an inflated *threshold* on the detector could mask a runaway.

2. **Lint**: cleaned everything introduced by this branch (unused imports, import ordering, `Optional` → `X | None` on the new field). The remaining ruff findings in the touched modules (e.g. the `MODEL_CONFIGS` protocol comment, `List` annotations) are pre-existing on `main` — happy to leave those for a separate cleanup unless you'd prefer them here.

Re-ran: 21/21 runaway+guard tests, 155 passed on the wider suite, `bun run typecheck` clean. Docstring coverage: the newly added functions all carry docstrings; the remaining gap is mostly pytest functions, which follow the existing style of `test_qwen_runaway_retry.py` (no docstrings) — let me know if you'd like them added anyway.

